PR #269, opened 2024-03-20
2.0.67-RELEASE Feature#200

laradelrio:
Challenge list translation in eng, esp, cat


laradelrio:
Cambio de dummy a end-point real

PR #275, opened 2024-03-21
RELEASE 2.0.70 - Feature#200

laradelrio:
Translation challenge list eng, es, ca from endpoint
(warning, the backend must still update the "cat" to "ca" and therefore cat  translation doesn't work yet)


laradelrio:
error

PR #292, opened 2024-04-08
2.7.0-RELEASE Feature#258

sacom14:
Tasks:
- Modify challenge.component.ts: delete margin-right 4rem for mobile view.
- Change width of nfbNavItem to 100% on challenge-info.component.html.
- We delete de button (send solution, etc.) from resource and related on challenge-header.component.
- We add class d-none for header buttons (send solution, etc.) of challenge-header.component and add the button on challenge-info.component.html (only for mobile view).


jonatanvicente:
* Es necesario revisar el resizing para resoluciones Desktop amplias :)

sacom14:
Resource sizing for desktop view changed

PR #398, opened 2024-07-02
Feature#300

Yul1b3th:


jonatanvicente:
This issue is wrong implemented, by next reasons:
* Auth service must be available to User service, but not otherwise. Auth service is generic, and doesn't need to know any details over user. Nevertheless, User needs access to all functionalities provided by Auth Service.
* Really, User Service do nothing in this code

PR #414, opened 2024-09-16
anadirme-contributors

adptCode:
anadir mi nombre a los contributors


jonatanvicente:
Please, add only files you need to resolve card issue.

PR #459, opened 2024-10-31
Feature#449

xapolin1979:


jonatanvicente:
Please, keep an eye to next:

* At starter.service.ts both functions orderBySortAscending and orderBySortAsDesceding are basically the same (only lines 48 and 66 are different). Please, refactor this: it should be only one function that order lists (it needs a parameter to know which order is prefered). This refactoring avoid lines 129-131 (it doesn't make sense).
* At starter.component.ts there are several subscriptions not closed. 
* This branch is not updated. Please, review this.

Issue #454, opened 2024-10-28
Config Management Refactoring


jonatanvicente:
Files that contain configuration are not identical (both environment.ts and environment.prod.ts). 
It's necessary to review them. In addition, many variable names are confused (including SSO prefix, for example).

PR #461, opened 2024-10-31
#454: Config Management Refactoring

PacMM79:


jonatanvicente:
There are some conflicts with the environment files. Keep an eye to this, keep in mind that other developer has changed several fields of that file.
Please, remember set the name to work branches using format **feature#[number]**. Thks in advance.

PR #503, opened 2025-01-23
Onboarding update

DaylinMB:


ivilarop (changes requested):
Only your name change should appear.

PR #522, opened 2025-02-10
Feature/519 remove resources

ivanetierra:
removed resource-card component
removed everything related to calling resources 
removed resource.service
add placeholder with "no resources" message


deuvede24 (approved):
looks ok

PR #530, opened 2025-02-20
Add static TimyMCE

ivilarop:


deuvede24 (approved):
Looks good! TinyMCE correctly copied into the NGINX container

PR #553, opened 2025-03-19
feature/111 Login Button in Header

mmartincasas:
✨ Added Login Button in Header & Improved Styles

- Added login button to the header for desktop and mobile.
- Improved styling.
- For now, the button is disabled when the session is active.


jhor94 (approved):
The changes and functionality are correct 😄

manuK1777 (approved):
The button works as intended, compared to the Figma design it could have a bit more left and right padding, but I don't think that's an issue. I don't see enhancements in the code.

ivilarop (approved):

PR #560, opened 2025-03-23
Feature/150 scrollbar design refiniement

GuillemBo:
- I traced the figma disign with a picture template in order to copy the "create new challenge" form everything has been done in 1024px screen sized. Notice that nav-bar is smaller in order to fit in the template

- A scroll bar has been added in the form.


mmartincasas (changes requested):
Here are a few things to keep in mind:
• The text in the side menu feels a bit small and hard to read.
• The navbar buttons could use a bit more bottom margin.
• The "Publish" button has a shadow and a color that don’t match the standard pink.
• I’d suggest removing the layout tracing images (home.jpg & home2.jpg) from the PR — they help when fine-tuning the layout, but aren’t really needed in development.
• After chatting with the designer, we’ll probably need to revisit font sizes in future iterations — they look a bit too small right now. Something around 14px would work better (if you have time to adjust now, great — if not, we can tackle it in the next sprint).

mmartincasas (approved):
Well done!

ivilarop (approved):

PR #567, opened 2025-04-02
Feature/193 UI adjustments user send solution

7ldsma:
This PR is related to the following part of user journey.
image

PR OBJECTIVE.
Apply style changes to align with the Figma design in two steps of the user solution submission journey:

1. Modal for submitting the final solution.

2. Solutions page displaying the personal solution and the solution provided by IT Academy.

Changes are illustrated in the following images:
1
1  Edición Modal
2
2  Challenge View

This PR also includes changes from previous branches that are currently under review.


src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -105,6 +105,14 @@ export class ChallengeHeaderComponent implements OnInit {
     })
     modalRef.componentInstance.idChallenge = this.idChallenge;
     modalRef.componentInstance.userId = this.userId;
+    modalRef.componentInstance.solutionAccepted.subscribe(() => {
```
martaag18, line 122: Just a quick suggestion on the commit message — "check for changes and set values" is a bit generic and could be more descriptive to reflect what’s actually being updated. 
Since this change connects the modal to the parent component and updates the UI after a solution is submitted, a more descriptive message could be:

`feat: update challenge header state when solution is submitted (via event emitter)`

That would make the commit easier to understand when reviewing the history.

martaag18 (changes requested):
Everything looks fine to me! 👌 The PR is well structured.

Just spotted a couple of things that might need sorting out:

🔴 **There are 2 failing tests**:
1. `SendSolutionModalComponent › should call submitSolution` — looks like the test is still using the old parameter order.
2. `ChallengeHeaderComponent › should open send solution modal` — the mock modal is missing the `solutionAccepted` event, so the `.subscribe()` call crashes.

📝 **Nice to have**:
It would be super helpful to include the Hito (epic) number from Figma in the screenshots or PR description. That way it’s easier for the team to match the implementation with the designs.

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
7ldsma: Thank you

7ldsma:
Tests fixed

martaag18 (approved):
Everything looks great now!

ivilarop (approved):

PR #588, opened 2025-05-05
Feature/266 update total likes counter

AlexandraBonetCanela:
Related User Story → This PR belongs to User Story 251 → Likes y bookmarks

What has been done in this PR? → Update total likes counter with Backend response when adding a challenge to favorites.

### DIFF
--- a/src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
+++ b/src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
@@ -155,13 +155,14 @@ export class ChallengeHeaderComponent implements OnInit {
       return
     }
     this.isFavorite = !this.isFavorite
-    this.favorites_count += this.isFavorite ? 1 : -1
     if (this.isFavorite) {
       this.challengeService.addToFavorites(this.idChallenge).subscribe({
         next: response => {
           console.log('Favorite added:', response)
+          this.favorites_count = response.timesFavorited;
         },
         error: () => {
+
         }
       })
     } else {
--- a/src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
+++ b/src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
@@ -157,9 +157,14 @@ export class ChallengeHeaderComponent implements OnInit {
     this.isFavorite = !this.isFavorite
     this.favorites_count += this.isFavorite ? 1 : -1
     if (this.isFavorite) {
-      this.removeFromFavorites()
+      this.challengeService.addToFavorites(this.idChallenge).subscribe({
+        next: () => {
+        },
+        error: () => {
+        }
+      })
     } else {
-      this.addToFavorites()
+    // TODO: Implementar removeFromFavorites en la PR correspondiente
     }
   }
 
--- a/src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
+++ b/src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
@@ -150,7 +150,10 @@ export class ChallengeHeaderComponent implements OnInit {
     this.openSendSolutionModal()
   }
 
-  toggleFavorite(): void {
+  toggleFavorite (): void {
+    if (!this.authService.isUserLoggedIn()) {
+      return
+    }
     if (this.isFavorite) {
       this.removeFromFavorites()
     } else {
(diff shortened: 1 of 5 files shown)


src/app/shared/components/challenge-card/challenge-card.component.ts
```diff
@@ -13,6 +13,7 @@ export class ChallengeCardComponent implements OnInit {
   private readonly starterService = inject(StarterService)
   private readonly translate = inject(TranslateService)
   private readonly challengeService = inject(ChallengeService)
+  private readonly authService = inject(AuthService)
```
martaag18, line 17: You should add import { AuthService } from 'src/app/auth/services/auth.service', otherwise the app won’t compile and the local environment won’t start properly.

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -150,11 +150,21 @@ export class ChallengeHeaderComponent implements OnInit {
     this.openSendSolutionModal()
   }
 
-  toggleFavorite(): void {
+  toggleFavorite (): void {
+    if (!this.authService.isUserLoggedIn()) {
+      return
+    }
+    this.isFavorite = !this.isFavorite
+    this.favorites_count += this.isFavorite ? 1 : -1
```
martaag18: This manual increment/decrement was only added as a temporary UI placeholder to show immediate feedback before we wired up the real backend response.
Now that we set `favorites_count` from the server’s `resp.timesFavorited`, we can remove this line.
this.favorites_count += this.isFavorite ? 1 : -1

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -150,11 +150,21 @@ export class ChallengeHeaderComponent implements OnInit {
...
     if (this.isFavorite) {
-      this.removeFromFavorites()
+      this.challengeService.addToFavorites(this.idChallenge).subscribe({
+        next: () => {
```
martaag18: We could add here the response from the backend:
 this.favorites_count = resp.timesFavorited;

martaag18 (changes requested):

src/app/shared/components/challenge-card/challenge-card.component.ts
```diff
@@ -49,7 +50,24 @@ export class ChallengeCardComponent implements OnInit {
 
   toggleFavorite (event: MouseEvent): void {
     event.stopPropagation()
+    if (!this.authService.isUserLoggedIn()) {
+      return
+    }
     this.isFavorite = !this.isFavorite
-    this.favorites_count = this.isFavorite ? this.favorites_count + 1 : Math.max(0, this.favorites_count - 1)
+    this.favorites_count += this.isFavorite ? 1 : -1
```
martaag18: Same as in ChallengeHeaderComponent, we should remove this line. 
It was only a temporary UI placeholder before we wired up the real backend response.

martaag18 (changes requested):

src/app/shared/components/challenge-card/challenge-card.component.ts
AlexandraBonetCanela: Done

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
AlexandraBonetCanela: I was debating whether to leave it or not, and decided to leave it so user got feedback at the moment but makes sense, changed.

src/app/shared/components/challenge-card/challenge-card.component.ts
AlexandraBonetCanela: Done

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
AlexandraBonetCanela: Done!

martaag18 (changes requested):
I've been testing the functionality, and it’s not working as expected. When I click the heart icon, the like counter increases by 2 instead of 1, so it looks like it's being updated twice.
image
I believe the issue is caused by mixing localStorage values with backend responses. In ngOnInit() we're loading isFavorite and favorites_count from localStorage, but then we update them based on the backend after clicking the heart icon. This creates inconsistent state and may cause the counter to behave incorrectly.

PR #591, opened 2025-05-07
Feature/327 remove like backend

martaag18:
This PR implements the logic to **remove a challenge from a user's favorites by making a DELETE request to the backend.**

The `removeFromFavorites(challengeId: string)` method has been updated in `ChallengeService` to:

- Replace the mock implementation using localStorage.
- Call the real backend endpoint: DELETE /challenge/challenges/{challengeId}/favorites.
- Include the required authentication headers (Bearer token).
- Return the response as an Observable of type `FavoriteResponse`.

Functionality has been tested manually and returns expected backend data.

**This task corresponds to user story 314 and completes task 327.**

### DIFF
--- a/src/app/services/challenge.service.ts
+++ b/src/app/services/challenge.service.ts
@@ -122,7 +122,6 @@ export class ChallengeService {
     );
   }
 
-  // Mocked version for frontend testing
   removeFromFavorites (challengeId: string): Observable<FavoriteResponse> {
     const headers = {
       'Content-Type': 'application/json',
--- a/src/app/services/challenge.service.ts
+++ b/src/app/services/challenge.service.ts
@@ -123,19 +123,21 @@ export class ChallengeService {
   }
 
   // Mocked version for frontend testing
-  removeFromFavorites(challengeId: string): Observable<FavoriteResponse> {
-
-    const currentCount = this.getMockFavoriteCount(challengeId);
-    
-    const mockResponse: FavoriteResponse = {
-      isFavorite: false,
-      timesFavorited: currentCount > 0 ? currentCount - 1 : 0
-    }
-    
-    localStorage.setItem(`is_favorite_${challengeId}`, 'false');
-    localStorage.setItem(`favorites_count_${challengeId}`, mockResponse.timesFavorited.toString());
-    
-    return of(mockResponse).pipe(delay(300));
+  removeFromFavorites (challengeId: string): Observable<FavoriteResponse> {
+    const headers = {
+      'Content-Type': 'application/json',
+      ...this.authService.getAuthHeaders()
+    };
+    const url = `${environment.BACKEND_ITA_CHALLENGE_BASE_URL}/challenge/challenges/${challengeId}/favorites`;
+    return this.http.delete<FavoriteResponse>(url, { headers }).pipe(
+      map(response => {
+        return response;
+      }),
+      catchError(error => {
+        console.error('Error removing from favorites:', error);
+        return of({ isFavorite: true, timesFavorited: 0 });
+      })
+    );
   }
 
   // Helper methods for mock implementation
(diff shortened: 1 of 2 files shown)


src/app/services/challenge.service.ts
```diff
@@ -123,19 +123,21 @@ export class ChallengeService {
...
-    localStorage.setItem(`is_favorite_${challengeId}`, 'false');
-    localStorage.setItem(`favorites_count_${challengeId}`, mockResponse.timesFavorited.toString());
-    
-    return of(mockResponse).pipe(delay(300));
+  removeFromFavorites (challengeId: string): Observable<FavoriteResponse> {
+    const headers = {
+      'Content-Type': 'application/json',
+      ...this.authService.getAuthHeaders()
+    };
+    const url = `${environment.BACKEND_ITA_CHALLENGE_BASE_URL}/challenge/challenges/${challengeId}/favorites`;
```
AlexMafaluno: there is a a path to /challenge/challenges/ in environment.ts

src/app/services/challenge.service.ts
```diff
@@ -123,19 +123,21 @@ export class ChallengeService {
...
+    return this.http.delete<FavoriteResponse>(url, { headers }).pipe(
+      map(response => {
+        return response;
+      }),
+      catchError(error => {
+        console.error('Error removing from favorites:', error);
```
AlexMafaluno, line 136: avoid consoles

martaag18: In this PR, I kept them just for debugging purposes, since I'm not closing the full lifecycle of the feature yet. I'll remove them as soon as I complete the flow in the upcoming PRs and before closing the user story.

AlexMafaluno (approved):
Hi Marta, everything meets the PR requirements. Approved.

src/app/services/challenge.service.ts
```diff
@@ -122,20 +122,21 @@ export class ChallengeService {
...
+      ...this.authService.getAuthHeaders()
+    };
+    const url = `${environment.BACKEND_ITA_CHALLENGE_BASE_URL}${environment.BACKEND_ALL_CHALLENGES_URL}/${challengeId}/favorites`;
+    return this.http.delete<FavoriteResponse>(url, { headers }).pipe(
+      map(response => {
+        return response;
+      }),
+      catchError(error => {
+        console.error('Error removing from favorites:', error);
+        return of({ isFavorite: true, timesFavorited: 0 });
```
ivilarop, line 137: Could you explain why this object is returned in the event of an error?
At first glance, it seems contradictory that, after failing to delete the favorite, isFavorite: true is returned, indicating that it is still marked as a favorite, but with timesFavorited: 0, which is inconsistent.

📌 It would be interesting to check if:

There is a functional reason behind it (e.g., preventing the frontend from breaking).

It makes sense to return a clearer alternative state (e.g., null, error: true, or propagating the error).

ivilarop: Could you explain why this object is returned in the event of an error?
At first glance, it seems contradictory that, after failing to delete the favorite, isFavorite: true is returned, indicating that it is still marked as a favorite, but with timesFavorited: 0, which is inconsistent.

📌 It would be interesting to check if:

There is a functional reason behind it (e.g., preventing the frontend from breaking).

It makes sense to return a clearer alternative state (e.g., null, error: true, or propagating the error).

ivilarop (changes requested):

src/app/services/challenge.service.ts
martaag18: I originally did it just to force the UI to stay in a valid state, avoiding any visual errors in the short term.
But you're right that it's cleaner to propagate the error and let the component handle it properly. I've now refactored the code to use throwError() instead, as you suggested.

ivilarop (approved):

ivilarop:
@martaag18 We must add this change to the changelog.

martaag18:
> @martaag18 We must add this change to the changelog.

As we agreed, I'm adding the **changelog** in the **last PR done of the UserStory - 329.**

PR #593, opened 2025-05-07
Feature/328 update favorite UI

martaag18:
This PR completes task 328, **part of user story 314** - "As a mentor, I can give/remove likes and see my liked challenges".

**Functional changes:**

- Updated the `toggleFavorite()` method in both `ChallengeCardComponent` and `ChallengeHeaderComponent` to call real backend endpoints.
- Removed local optimistic updates (`isFavorite`, `favorites_count`) in favor of backend response.
- Connected both `addToFavorites` and `removeFromFavorites` directly from `ChallengeService`.
- Deleted unused internal methods in `ChallengeHeaderComponent` that previously handled favorite logic.

 **How to Test It**

1. Log in as a mentor.
2. Go to the challenge list and click the heart icon (toggle favorite).
3. Verify:
   - When liking a challenge: a POST request is made to the backend.
   - When unliking: a DELETE request is made.
   - The heart icon updates correctly (filled/empty).
   - The favorites counter updates based on the backend response.

4. Navigate to the challenge detail page:
   - Use the heart icon in the header.
   - Verify the same behavior (add/remove favorites) works as expected.

### DIFF
--- a/src/app/models/favorite-response.interface.ts
+++ b/src/app/models/favorite-response.interface.ts
@@ -1,4 +1,4 @@
 export interface FavoriteResponse {
-  isFavorite: boolean
+  favorite: boolean
   timesFavorited: number
 }
--- a/src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
+++ b/src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
@@ -188,22 +188,6 @@ export class ChallengeHeaderComponent implements OnInit {
     }
   }
 
-  private addToFavorites(): void {
-    this.challengeService.addToFavorites(this.idChallenge).subscribe(response => {
-      this.isFavorite = response.isFavorite
-      this.favorites_count = response.timesFavorited
-      this.favoritesUpdated.emit(this.favorites_count)
-    })
-  }
-
-  private removeFromFavorites(): void {
-    this.challengeService.removeFromFavorites(this.idChallenge).subscribe(response => {
-      this.isFavorite = response.isFavorite
-      this.favorites_count = response.timesFavorited
-      this.favoritesUpdated.emit(this.favorites_count)
-    })
-  }
-
   private checkFavoriteStatus(): void {
     const isFavorited = localStorage.getItem(`is_favorite_${this.idChallenge}`);
     this.isFavorite = isFavorited === 'true';
--- a/src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
+++ b/src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
@@ -164,7 +164,7 @@ export class ChallengeHeaderComponent implements OnInit {
     if (this.isFavorite) {
       this.challengeService.removeFromFavorites(this.idChallenge).subscribe({
         next: response => {
-          this.isFavorite = response.isFavorite
+          this.isFavorite = false
           this.favorites_count = response.timesFavorited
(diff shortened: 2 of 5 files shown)


src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -161,44 +159,30 @@ export class ChallengeHeaderComponent implements OnInit {
...
-    this.isFavorite = !this.isFavorite
-    this.favorites_count += this.isFavorite ? 1 : -1
     if (this.isFavorite) {
-      this.challengeService.addToFavorites(this.idChallenge).subscribe({
-        next: () => {
+      this.challengeService.removeFromFavorites(this.idChallenge).subscribe({
+        next: response => {
+          this.isFavorite = response.favorite
+          this.favorites_count = response.timesFavorited
+          console.log('Favorite removed:', response)
```
AlexMafaluno: remove consoles

AlexMafaluno (approved):
Hi, the heart icon now updates correctly, reflecting the number of times the challenge has been favourited. This part of the code is working well.

src/app/services/challenge.service.ts
```diff
@@ -117,25 +117,26 @@ export class ChallengeService {
...
+      ...this.authService.getAuthHeaders()
+    };
+    const url = `${environment.BACKEND_ITA_CHALLENGE_BASE_URL}/challenge/challenges/${challengeId}/favorites`;
+    return this.http.delete<FavoriteResponse>(url, { headers }).pipe(
+      map(response => {
+        return response;
+      }),
+      catchError(error => {
+        console.error('Error removing from favorites:', error);
+        return of({ favorite: true, timesFavorited: 0 });
```
ivilarop: Could you explain why this object is returned in the event of an error?
At first glance, it seems contradictory that, after failing to delete the favorite, isFavorite: true is returned, indicating that it is still marked as a favorite, but with timesFavorited: 0, which is inconsistent.

📌 It would be interesting to check if:

There is a functional reason behind it (e.g., preventing the frontend from breaking).

It makes sense to return a clearer alternative state (e.g., null, error: true, or propagating the error).

src/app/models/favorite-response.interface.ts
```diff
@@ -1,4 +1,4 @@
 export interface FavoriteResponse {
-  isFavorite: boolean
+  favorite: boolean
```
ivilarop, line 2: I see that isFavorite has been changed to favorite in the FavoriteResponse interface. Is there a specific reason for this change?

Is it due to a tweak to the backend response?

Or is it just for simplicity?

Keep in mind that, if there's no real need, these types of naming changes generate unnecessary refactorings that affect multiple files and can make it difficult to trace the code history.

ivilarop (changes requested):

src/app/services/challenge.service.ts
martaag18: I implemented the change from my last PR 327. 
So now I am propagating the error.

src/app/models/favorite-response.interface.ts
martaag18: image

I changed isFavorite to favorite in the interface because that's how the backend actually returns the field (you can see it in the screenshot I added above).
I wouldn’t normally rename things unless necessary, but in this case, it was just to reflect the real structure from the backend and avoid future confusion

ivilarop (approved):

PR #608, opened 2025-05-12
Feature/357 progress filter UI

AlexMafaluno:
This PR completes **task 357,** part of user story **312 - "As a mentor, I can filter the challenges”.**

The goal of this PR is to allow users with the ADMIN role to view and use the "Progress" section within the challenge filters, which was previously restricted. **This requirement needs only visibility, without Functionality.**

Functional changes:

Updated the logic that controls visibility of the Progress filter section:

- Previously, it was hidden for users with the ADMIN role.

- Now, it is visible and fully functional for both authenticated users and admins.

- Updated unit tests to reflect this role-based behavior, ensuring ADMIN users also trigger the expected UI rendering.

### DIFF
--- a/src/app/modules/starter/components/starter-filters/starter-filters.component.html
+++ b/src/app/modules/starter/components/starter-filters/starter-filters.component.html
@@ -35,7 +35,7 @@
     </div>
   </div>
   
-  <div class="form" formGroupName="progress">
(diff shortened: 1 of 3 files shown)


martaag18 (approved):
It looks good! 
I tested locally and as a mentor, I can view the sector "progress".

ivilarop (approved):

PR #615, opened 2025-05-13
Feature/360 popularity filer logic v02

bertapasamontes:
This PR completes task **360**, part of user story 313 - "General layout review for mentor screens".

Popularity filter is active when there is not any other active filter in localstorage when user recharge the page.

Files changed: 
- starter.component.ts
- starter.service.ts

### DIFF
--- a/src/app/modules/starter/components/starter/starter.component.ts
+++ b/src/app/modules/starter/components/starter/starter.component.ts
@@ -51,7 +51,6 @@ export class StarterComponent implements OnInit {
     this.userRoleSubs$ = this._authService.getUserRole().subscribe(role => {
       this.isAdmin = role === 'ADMIN'
     })
-<<<<<<< HEAD
     if (this._authService.isUserLoggedIn()) {
       this._authService.getUserId().subscribe(userId => {
         if (userId !== null && userId !== '') {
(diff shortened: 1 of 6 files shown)


src/app/modules/starter/components/starter/starter.component.ts
```diff
@@ -52,6 +52,15 @@ export class StarterComponent implements OnInit {
     this.userRoleSubs$ = this._authService.getUserRole().subscribe(role => {
       this.isAdmin = role === 'ADMIN'
     })
+
+    const savedSort = localStorage.getItem('sortBy')
```
AlexMafaluno: Hi Berta, I have a question. Is there a specific reason we're retrieving the sort option from localStorage?
If we always want the app to initialize with popularity as the default, wouldn’t it be simpler and cleaner to just set sortBy: string = 'popularity' directly?
This way, we can avoid relying on localStorage and reduce unnecessary logic.

AlexMafaluno (changes requested):
Hi, Berta, i put some comments below.

src/app/modules/starter/components/starter/starter.component.ts
bertapasamontes: The reason for saving it in localStorage was to make it easier for the user to navigate by keeping their most recently used filter but we can remove it if needed.

AlexMafaluno: Ok Berta,Thanks! For a future refactor, it would be great to take this into consideration. For this PR, however, it's all set.

AlexMafaluno (approved):

ivilarop (approved):

PR #621, opened 2025-05-21
feat(user): assign profile users photo (#433)

bertapasamontes:
### Description
This PR completes task **433**, part of user story 388 - "As a user, I can see my role and be able to modify it in the profile drop-down menu.".

A GET request was made to retrieve the user's profile picture URL, and it was added to the navbar, replacing the profile icon.
In addition, a refactor of shared styles was carried out.
Reusable style rules were extracted and moved into the _global.scss file to promote consistency and reduce duplication across components.

### How to test it:
- Sign in with github
- Check if your github profile image is displayed in the menu.

### Before and after comparison
**Before:**
image
image

**After (desktop):**
image

**After (mobile):**
image

### Files changed:
- environmet
- auth.service
- desktop-nav
- mobile-nav
- _global.scss

### Auto self-check:

- [x] I tested my changes locally and verified they work as expected
- [ ] The PR has a clear and focused purpose (only one feature, fix or refactor)
- [x] Title, description, and changelog (if needed) are filled in correctly
- [x] There are no leftover merge conflicts or unrelated changes from previous branches
- [ ] All automated checks (like SonarCloud) have passed
- [x] I responded to all previous review comments and only resolved those I truly addressed

### DIFF
--- a/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
+++ b/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
@@ -20,7 +20,7 @@
       }
       @else {
         <button class="user userWithPhoto" (click)="toggleDropdown()">
-          <img [src]="userPhoto" alt="User" class="img-fluid profilePhoto" aria-label="User menu">
+          <img [src]="userPhoto" alt="User" class="img-fluid" aria-label="User menu">
         </button>
       }
       <div class="dropdown" [ngClass]="{ 'show': dropdownOpen }">
--- a/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
+++ b/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
@@ -14,7 +14,7 @@
   @if(isLoggedIn){
     <div class="user-container">
       <button [ngClass]="{'user':true, 'userWithPhoto':!userPhoto}" (click)="toggleDropdown()">
-        <img src="userPhoto ? userPhoto : assets/img/icon/user.svg" alt="User" [ngClass]="{'img-fluid': !userPhoto}">
+        <img src="userPhoto ? userPhoto : assets/img/icon/user.svg" [alt]="'components.header.userPhoto' | translate" [ngClass]="{'img-fluid': !userPhoto}">
       </button>
       <div class="dropdown" [ngClass]="{ 'show': dropdownOpen }">
           <a>{{user}}</a>
--- a/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
+++ b/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
@@ -13,16 +13,6 @@
   <!-- Login -->
   @if(isLoggedIn){
     <div class="user-container">
-       <!-- @if (!userPhoto) {
-        <button class="user" (click)="toggleDropdown()">
-          <img src="assets/img/icon/user.svg" alt="User" style="border: none; padding: 0; background: none;">
-        </button>
-      }
-      @else {
-        <button class="user userWithPhoto" (click)="toggleDropdown()">
-          <img [src]="userPhoto" alt="User" class="img-fluid" aria-label="User menu">
-        </button>
-      } -->
       <button [ngClass]="{'user':true, 'userWithPhoto':!userPhoto}" (click)="toggleDropdown()">
         <img src="userPhoto ? userPhoto : assets/img/icon/user.svg" alt="User" [ngClass]="{'img-fluid': !userPhoto}">
       </button>
--- a/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
+++ b/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
@@ -20,7 +20,7 @@
       }
       @else {
         <button class="user userWithPhoto" (click)="toggleDropdown()">
-          <img [src]="userPhoto" alt="User" class="img-fluid profilePhoto">
+          <img [src]="userPhoto" alt="User" class="img-fluid profilePhoto" aria-label="User menu">
         </button>
       }
       <div class="dropdown" [ngClass]="{ 'show': dropdownOpen }">
--- a/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
+++ b/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
@@ -13,9 +13,16 @@
   <!-- Login -->
   @if(isLoggedIn){
     <div class="user-container">
-      <button class="user" (click)="toggleDropdown()">
-        <img src="assets/img/icon/user.svg" alt="User" style="border: none; padding: 0; background: none;">
-      </button>
+       @if (!userPhoto) {
+        <button class="user" (click)="toggleDropdown()">
+          <img src="assets/img/icon/user.svg" alt="User" style="border: none; padding: 0; background: none;">
+        </button>
+      }
+      @else {
+        <button class="userWithPhoto" (click)="toggleDropdown()">
+          <img [src]="userPhoto" alt="User" class="img-fluid profilePhoto" style="border: none; padding: 0; background: none;">
+        </button>
+      }
       <div class="dropdown" [ngClass]="{ 'show': dropdownOpen }">
           <a>{{user}}</a>
           <a>
--- a/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
+++ b/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
@@ -20,7 +20,7 @@
       }
       @else {
         <button class="userWithPhoto" (click)="toggleDropdown()">
-          <img [src]="userPhoto" alt="User" class="img-fluid profilePhoto" style="border: none; padding: 0; background: none;">
+          <img [src]="userPhoto" alt="User" class="img-fluid profilePhoto">
         </button>
       }
       <div class="dropdown" [ngClass]="{ 'show': dropdownOpen }">
--- a/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
+++ b/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
@@ -13,8 +13,8 @@
   <!-- Login -->
   @if(isLoggedIn){
     <div class="user-container">
-      <button [ngClass]="{'user':true, 'userWithPhoto':!userPhoto}" (click)="toggleDropdown()">
-        <img src="userPhoto ? userPhoto : assets/img/icon/user.svg" [alt]="'components.header.userPhoto' | translate" [ngClass]="{'img-fluid': !userPhoto}">
+      <button [ngClass]="{'user':true, 'userWithPhoto':userPhoto !=''}" (click)="toggleDropdown()">
+        <img [src]="userPhoto ? userPhoto : 'assets/img/icon/user.svg'" [alt]="'components.header.userPhoto' | translate" [ngClass]="{'img-fluid': userPhoto !=''}">
       </button>
       <div class="dropdown" [ngClass]="{ 'show': dropdownOpen }">
           <a>{{user}}</a>
--- a/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
+++ b/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
@@ -13,7 +13,7 @@
   <!-- Login -->
   @if(isLoggedIn){
     <div class="user-container">
-       @if (!userPhoto) {
+       <!-- @if (!userPhoto) {
         <button class="user" (click)="toggleDropdown()">
           <img src="assets/img/icon/user.svg" alt="User" style="border: none; padding: 0; background: none;">
         </button>
@@ -22,7 +22,10 @@
         <button class="user userWithPhoto" (click)="toggleDropdown()">
           <img [src]="userPhoto" alt="User" class="img-fluid" aria-label="User menu">
         </button>
-      }
+      } -->
+      <button [ngClass]="{'user':true, 'userWithPhoto':!userPhoto}" (click)="toggleDropdown()">
+        <img src="userPhoto ? userPhoto : assets/img/icon/user.svg" alt="User" [ngClass]="{'img-fluid': !userPhoto}">
+      </button>
       <div class="dropdown" [ngClass]="{ 'show': dropdownOpen }">
           <a>{{user}}</a>
           <a>
--- a/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
+++ b/src/app/core/layout/header/desktop-nav/desktop-nav.component.html
@@ -19,7 +19,7 @@
         </button>
       }
       @else {
-        <button class="userWithPhoto" (click)="toggleDropdown()">
+        <button class="user userWithPhoto" (click)="toggleDropdown()">
           <img [src]="userPhoto" alt="User" class="img-fluid profilePhoto">
         </button>
       }
--- a/src/app/core/layout/header/desktop-nav/desktop-nav.component.scss
+++ b/src/app/core/layout/header/desktop-nav/desktop-nav.component.scss
@@ -89,14 +89,25 @@
 
   .user {
     border: unset;
-    width: 48px;
-    height: 41px;
     background: $white;
     border-radius: 10px;
     width: 45px;
     height: 38px;
     position: relative;
     display: inline-block;
+  }
+
+  .userWithPhoto {
+    border-radius: 50%;
+    border: unset;
+    width: 50px;
+    height: 50px;
+    position: relative;
+    display: inline-block;
+  }
+
+  .profilePhoto{
+    border-radius: 50px;
   }
 
   .dropdown {
--- a/src/app/core/layout/header/desktop-nav/desktop-nav.component.scss
+++ b/src/app/core/layout/header/desktop-nav/desktop-nav.component.scss
@@ -87,30 +87,6 @@
   display: inline-block;
 }
 
-  .user {
-    border: unset;
-    background: $white;
-    border-radius: 10px;
-    width: 45px;
-    height: 38px;
-    position: relative;
-    display: inline-block;
-  }
-
-   .userWithPhoto {
-    border-radius: 50%;
-    width: 40px;
-    height: 40px;
-    position: relative;
-    display: inline-block;
-    padding: 0;
-    overflow: hidden;
-   }
-
-  .profilePhoto{
-    border-radius: 50px;
-  }
-
   .dropdown {
     display: none;
     position: absolute;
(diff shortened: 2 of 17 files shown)


src/app/core/layout/header/desktop-nav/desktop-nav.component.scss
```diff
@@ -99,6 +97,19 @@
     display: inline-block;
   }
 
+  .userWithPhoto {
```
basantades: I'd suggest updating the .userWithPhoto class to the following:

.userWithPhoto {
  border-radius: 50%;
  width: 40px;
  height: 40px;
  position: relative;
  display: inline-block;
  padding: 0;
  overflow: hidden;
}


border: unset seems unnecessary here, so I removed it.

Updated the size to 40px to better match the Figma design and avoid the top cropping issue (some of the container/layout CSS is causing unexpected shifts).

Added padding: 0 to prevent a strange white border caused by inherited padding from a parent element.

Added overflow: hidden so the image doesn't need a separate class to apply the border radius. This way, you can control the circular shape from a single place.

Let me know if that works for you!

src/app/core/layout/header/mobile-nav/mobile-nav.component.scss
```diff
@@ -109,6 +109,19 @@ $navbar-toggler-transition: none;
       position: relative;
       display: inline-block;
     }
+
+    .userWithPhoto {
```
basantades: Same as in the other CSS

I'd suggest updating the .userWithPhoto class to the following:

.userWithPhoto {
border-radius: 50%;
width: 40px;
height: 40px;
position: relative;
display: inline-block;
padding: 0;
overflow: hidden;
}

border: unset seems unnecessary here, so I removed it.

Updated the size to 40px to better match the Figma design and avoid the top cropping issue (some of the container/layout CSS is causing unexpected shifts).

Added padding: 0 to prevent a strange white border caused by inherited padding from a parent element.

Added overflow: hidden so the image doesn't need a separate class to apply the border radius. This way, you can control the circular shape from a single place.

Let me know if that works for you!

src/app/core/layout/header/desktop-nav/desktop-nav.component.html
```diff
@@ -13,9 +13,16 @@
   <!-- Login -->
   @if(isLoggedIn){
     <div class="user-container">
-      <button class="user" (click)="toggleDropdown()">
-        <img src="assets/img/icon/user.svg" alt="User" style="border: none; padding: 0; background: none;">
-      </button>
+       @if (!userPhoto) {
+        <button class="user" (click)="toggleDropdown()">
```
basantades: Please consider adding aria-label="User menu" to the <button> element to improve accessibility for screen readers.

src/app/core/layout/header/desktop-nav/desktop-nav.component.html
```diff
@@ -13,9 +13,16 @@
...
+          <img src="assets/img/icon/user.svg" alt="User" style="border: none; padding: 0; background: none;">
+        </button>
+      }
+      @else {
+        <button class="user userWithPhoto" (click)="toggleDropdown()">
```
basantades: Please consider adding aria-label="User menu" to the <button> element to improve accessibility for screen readers.

src/app/core/layout/header/mobile-nav/mobile-nav.component.html
```diff
@@ -24,9 +24,16 @@
       <!-- Login -->
       @if(isLoggedIn){
         <div class="user-container-mobile">
-          <button class="user-mobile" (click)="toggleDropdown()">
-            <img src="assets/img/icon/user.svg" alt="User" style="border: none; padding: 0; background: none;">
-          </button>
+          @if (!userPhoto) {
+            <button class="user-mobile" (click)="toggleDropdown()">
```
basantades: Please consider adding aria-label="User menu" to the <button> element to improve accessibility for screen readers.

src/app/core/layout/header/mobile-nav/mobile-nav.component.html
```diff
@@ -24,9 +24,16 @@
...
+              <img src="assets/img/icon/user.svg" alt="User" style="border: none; padding: 0; background: none;">
+            </button>
+          }
+          @else {
+            <button class="user-mobile userWithPhoto" (click)="toggleDropdown()">
```
basantades: Please consider adding aria-label="User menu" to the <button> element to improve accessibility for screen readers.

basantades:
Hi! I added a few inline suggestions related to the CSS and accessibility, such as simplifying the avatar styling and improving screen reader support with aria-label.

Here's a quick before/after comparison to show the visual effect of the CSS adjustments:
image
image

Additionally, it might be worth considering extracting the user profile photo into a separate component (e.g. UserPhotoComponent) since it's used in multiple places and could help keep the layout components cleaner and more focused.

Let me know what you think!

bertapasamontes:
> Hi! I added a few inline suggestions related to the CSS and accessibility, such as simplifying the avatar styling and improving screen reader support with aria-label.
> 
> Here's a quick before/after comparison to show the visual effect of the CSS adjustments: image image
> 
> Additionally, it might be worth considering extracting the user profile photo into a separate component (e.g. UserPhotoComponent) since it's used in multiple places and could help keep the layout components cleaner and more focused.
> 
> Let me know what you think!

Thanks, Alberto, for your suggestions. I've already implemented the style changes. Regarding the new component, I believe it would be more efficient to refactor the entire menu into a new component, rather than creating one solely for the user's photo. What do you think?

basantades:
> Thanks, Alberto, for your suggestions. I've already implemented the style changes. Regarding the new component, I believe it would be more efficient to refactor the entire menu into a new component, rather than creating one solely for the user's photo. What do you think?

Yes, that sounds like a good idea. Still, even when refactoring the menu, it might be a good idea to call a small avatar component from within it 🙂.

basantades (changes requested):

src/app/core/layout/header/desktop-nav/desktop-nav.component.html
bertapasamontes: > Please consider adding aria-label="User menu" to the element to improve accessibility for screen readers.

done :)

src/app/core/layout/header/desktop-nav/desktop-nav.component.html
bertapasamontes: > Please consider adding aria-label="User menu" to the element to improve accessibility for screen readers.

done :)

src/app/core/layout/header/mobile-nav/mobile-nav.component.html
bertapasamontes: > Please consider adding aria-label="User menu" to the element to improve accessibility for screen readers.

done :)

src/app/core/layout/header/mobile-nav/mobile-nav.component.html
bertapasamontes: done :)

src/app/core/layout/header/mobile-nav/mobile-nav.component.scss
bertapasamontes: > Same as in the other CSS
> 
> I'd suggest updating the .userWithPhoto class to the following:
> 
> .userWithPhoto { border-radius: 50%; width: 40px; height: 40px; position: relative; display: inline-block; padding: 0; overflow: hidden; }
> 
> border: unset seems unnecessary here, so I removed it.
> 
> Updated the size to 40px to better match the Figma design and avoid the top cropping issue (some of the container/layout CSS is causing unexpected shifts).
> 
> Added padding: 0 to prevent a strange white border caused by inherited padding from a parent element.
> 
> Added overflow: hidden so the image doesn't need a separate class to apply the border radius. This way, you can control the circular shape from a single place.
> 
> Let me know if that works for you!

done :)

src/app/core/layout/header/desktop-nav/desktop-nav.component.scss
bertapasamontes: > I'd suggest updating the .userWithPhoto class to the following:
> 
> .userWithPhoto { border-radius: 50%; width: 40px; height: 40px; position: relative; display: inline-block; padding: 0; overflow: hidden; }
> 
> border: unset seems unnecessary here, so I removed it.
> 
> Updated the size to 40px to better match the Figma design and avoid the top cropping issue (some of the container/layout CSS is causing unexpected shifts).
> 
> Added padding: 0 to prevent a strange white border caused by inherited padding from a parent element.
> 
> Added overflow: hidden so the image doesn't need a separate class to apply the border radius. This way, you can control the circular shape from a single place.
> 
> Let me know if that works for you!

done :)

basantades (approved):
All good on my side.

src/app/core/layout/header/desktop-nav/desktop-nav.component.html
```diff
@@ -13,9 +13,16 @@
   <!-- Login -->
   @if(isLoggedIn){
     <div class="user-container">
-      <button class="user" (click)="toggleDropdown()">
-        <img src="assets/img/icon/user.svg" alt="User" style="border: none; padding: 0; background: none;">
-      </button>
+       @if (!userPhoto) {
+        <button class="user" (click)="toggleDropdown()">
+          <img src="assets/img/icon/user.svg" alt="User" style="border: none; padding: 0; background: none;">
```
ivilarop: The image's alt attribute is statically set to "User." To maintain consistency with the application's internationalization system and improve accessibility, would it be advisable to translate it?

src/app/core/layout/header/desktop-nav/desktop-nav.component.html
```diff
@@ -13,9 +13,16 @@
...
+        </button>
+      }
+      @else {
+        <button class="user userWithPhoto" (click)="toggleDropdown()">
+          <img [src]="userPhoto" alt="User" class="img-fluid" aria-label="User menu">
```
ivilarop: The image's alt attribute is statically set to "User." To maintain consistency with the application's internationalization system and improve accessibility, would it be advisable to translate it?

src/app/core/layout/header/desktop-nav/desktop-nav.component.html
```diff
@@ -13,9 +13,16 @@
   <!-- Login -->
   @if(isLoggedIn){
     <div class="user-container">
-      <button class="user" (click)="toggleDropdown()">
-        <img src="assets/img/icon/user.svg" alt="User" style="border: none; padding: 0; background: none;">
-      </button>
+       @if (!userPhoto) {
```
ivilarop: The if/else blocks are similar. Both buttons do the same thing (toggleDropdown()), the only thing that changes is: The class (user vs user userWithPhoto) and The content (different img). Recommendation: extract the dynamic class and conditional content with ngClass

src/app/core/layout/header/desktop-nav/desktop-nav.component.ts
```diff
@@ -32,8 +33,9 @@ export class DesktopNavComponent implements OnInit, OnDestroy{
     this._authService.getUsername().subscribe((username) => {
       this.user = username;
     });
-    this._authService.getUserRole().subscribe((userRole) => {
-      this.currentRole = userRole
+
+    this._authService.getUserPhoto().subscribe((userPhoto) => {
```
ivilarop, line 43: Tip: Since each call (getUsername, getUserPhoto, updateUserRoleAndUserNameFromToken) is handled differently and there's no way to unify it in a single endpoint, it's justified to maintain separate subscriptions. Still, we could improve readability by extracting that logic into a loadUserBasicData()-like method.

This helps keep ngOnInit() clean and separate by responsibility, especially if we add error handling or loading in the future.

src/app/core/layout/header/mobile-nav/mobile-nav.component.html
```diff
@@ -24,9 +24,16 @@
       <!-- Login -->
       @if(isLoggedIn){
         <div class="user-container-mobile">
-          <button class="user-mobile" (click)="toggleDropdown()">
-            <img src="assets/img/icon/user.svg" alt="User" style="border: none; padding: 0; background: none;">
-          </button>
+          @if (!userPhoto) {
```
ivilarop: same comments that desktop

src/app/core/layout/header/mobile-nav/mobile-nav.component.ts
```diff
@@ -31,8 +31,8 @@ export class MobileNavComponent implements OnInit, OnDestroy{
       this.user = username;
     });
 
-    this._authService.getUserRole().subscribe((userRole) => {
-      this.currentRole = userRole
+    this._authService.getUserPhoto().subscribe((userPhoto) => {
```
ivilarop, line 41: same comments that desktop

src/app/services/auth.service.spec.ts
```diff
@@ -164,4 +169,22 @@ describe('AuthService', () => {
     expect(headers.Authorization).toBe('');
   });
 
+  it('should return user photo URL from GitHub', fakeAsync(() => {
+    const photoUrl = 'https://github.com/avatar.jpg'
+    const mockResponse = { avatar_url: photoUrl };
+
+    // Asigna un username al servicio usando spy o cambia visibilidad de la propiedad si necesario
```
ivilarop: necessary?

src/app/services/auth.service.ts
```diff
@@ -51,6 +70,9 @@ export class AuthService {
       this.usernameSubject.next(this.username);
       const userId = decodedToken?.uuid ?? null;
       this.userIdSubject.next(userId);
+      this.getUserPhoto().subscribe(photo => {
```
ivilarop: Suggestion: Subscribing to getUserPhoto() inside updateUserRoleAndUserNameFromToken() seems unnecessary, since the getUserPhoto() method itself internally assigns the value to this.userPhoto. We could remove this subscription, since nothing useful is being done with the received value. If we want to expose this data to other components in the future, we could also emit it through a Subject or BehaviorSubject.

src/app/services/auth.service.ts
```diff
@@ -22,12 +22,31 @@ export class AuthService {
...
 
+  public userPhoto: string | null = ''
+
   constructor(private http: HttpClient, private router: Router, private toastr: ToastrService, private translate: TranslateService) {
     this.translate.addLangs(['en', 'es', 'ca'])
     this.translate.setDefaultLang('ca')
     this.translate.use('ca')
   }
 
+  getUserPhoto(): Observable<string> {
```
ivilarop, line 33: Currently, a BehaviorSubject is used, and a manual subscribe() is performed within the service itself. Although functional, this is not the most idiomatic way to work with RxJS.

Additionally, the 'error' string is emitted in the event of a failure, which can mix types in the observable (URL vs. 'error'), making subsequent handling from the component difficult.

It would be interesting to investigate how to improve this pattern by using RxJS operators such as pipe, map, tap, and catchError, while maintaining the declarative flow and avoiding subscribing directly from the service.

ivilarop (changes requested):

src/environments/environment.ts
ivilarop, line 1: and shouldn't it be added to the environment.prod to work when deployed to development?

ivilarop (changes requested):

src/environments/environment.ts
bertapasamontes: > and shouldn't it be added to the environment.prod to work when deployed to development?

done

src/app/core/layout/header/desktop-nav/desktop-nav.component.html
bertapasamontes: > The image's alt attribute is statically set to "User." To maintain consistency with the application's internationalization system and improve accessibility, would it be advisable to translate it?

done

src/app/core/layout/header/desktop-nav/desktop-nav.component.html
bertapasamontes: > The image's alt attribute is statically set to "User." To maintain consistency with the application's internationalization system and improve accessibility, would it be advisable to translate it?

done

src/app/core/layout/header/desktop-nav/desktop-nav.component.html
bertapasamontes: > The if/else blocks are similar. Both buttons do the same thing (toggleDropdown()), the only thing that changes is: The class (user vs user userWithPhoto) and The content (different img). Recommendation: extract the dynamic class and conditional content with ngClass

done

src/app/core/layout/header/desktop-nav/desktop-nav.component.ts
bertapasamontes: > Tip: Since each call (getUsername, getUserPhoto, updateUserRoleAndUserNameFromToken) is handled differently and there's no way to unify it in a single endpoint, it's justified to maintain separate subscriptions. Still, we could improve readability by extracting that logic into a loadUserBasicData()-like method.
> 
> This helps keep ngOnInit() clean and separate by responsibility, especially if we add error handling or loading in the future.

done

src/app/core/layout/header/mobile-nav/mobile-nav.component.html
bertapasamontes: > same comments that desktop

done

src/app/core/layout/header/mobile-nav/mobile-nav.component.ts
bertapasamontes: > same comments that desktop

done

src/app/services/auth.service.spec.ts
bertapasamontes: > necessary?

done

src/app/services/auth.service.ts
bertapasamontes: > Suggestion: Subscribing to getUserPhoto() inside updateUserRoleAndUserNameFromToken() seems unnecessary, since the getUserPhoto() method itself internally assigns the value to this.userPhoto. We could remove this subscription, since nothing useful is being done with the received value. If we want to expose this data to other components in the future, we could also emit it through a Subject or BehaviorSubject.

done

src/app/services/auth.service.ts
bertapasamontes: > Currently, a BehaviorSubject is used, and a manual subscribe() is performed within the service itself. Although functional, this is not the most idiomatic way to work with RxJS.
> 
> Additionally, the 'error' string is emitted in the event of a failure, which can mix types in the observable (URL vs. 'error'), making subsequent handling from the component difficult.
> 
> It would be interesting to investigate how to improve this pattern by using RxJS operators such as pipe, map, tap, and catchError, while maintaining the declarative flow and avoiding subscribing directly from the service.

done

ivilarop (approved):

PR #651, opened 2025-06-18
feat: hide 'Start Challenge' button for ADMIN users (Taiga #527)

Silvitui:
This PR completes task **527**, part of user story **516** "As a mentor, I want the navigation for viewing challenge details and creating a challenge to match the Figma design."

---

### Purpose 

This PR **hides the "Start Challenge" button for ADMIN users (mentors)** in the challenge view. Since mentors don't need to participate in the challenge like students do, this UI element should be hidden to reduce confusion and prevent unnecessary actions.

---

### Changes included 

- Updated `challenge-header.component.html` to conditionally hide the "Start Challenge" button if the user role is `'ADMIN'`.
- Added a new unit test in `challenge-header.component.spec.ts` to ensure the button is **not shown** to mentors (`userRole = 'ADMIN'`).
- Also kept test coverage for `USER` role showing the button.

---

### Before and after

**Before:**
- All users, including mentors, saw the "Start Challenge" button.

**After:**
- Mentors (ADMIN role) no longer see the button.

---

### Files Changed

- `challenge-header.component.html`
- `challenge-header.component.spec.ts`
- `package.json`
- `package-lock.json`
- `CHANGELOG.md`

---

### PR Author Self-check

- [x] I tested my changes locally and verified they work as expected.
- [x] The PR has a clear and focused purpose (only one feature or fix).
- [x] Title, description, and changelog are filled in correctly.
- [x] There are no leftover merge conflicts or unrelated changes.
- [x] All automated checks (like SonarCloud) pass.
- [x] I responded to all review comments and only resolved what was addressed.

PR #653, opened 2025-06-25
chore(challenge): review mentor workflow and layout consistency (#529)

Silvitui:
This PR completes task **529**, part of user story **516**: _"Fix console errors and verify if there are any related issues in the Network tab."_

### 🧩 Context
This PR focuses on reviewing the mentor workflow to ensure that the experience is error-free and visually consistent. It includes verifying that no console or network errors occur and that the layout across all mentor-related views is clean and aligned with the Figma design.

### ✅ What was done

- Reviewed the full mentor experience:
  - Logging in as a mentor (ADMIN role)
  - Navigating to the list of challenges
  - Viewing a challenge detail
  - Creating a new challenge
- Verified:
  - No console errors
  - No failed network requests
  - Layout and spacing are correct, with no overlapping or misaligned elements

### 🎯 Goal
Ensure the mentor-facing challenge views are stable, visually clean, and work as expected without console or network issues.

### 🧪 How to test
1. Log in as a mentor (ADMIN role)
2. Go to:
   - `/challenges`
   - `/challenges/:id`
   - `/challenges/new`
3. Check:
   - No console or network errors
   - Layout is consistent, no excessive spacing or overlaps
   - Flow is smooth and matches the expected design


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [itachallenge-frontend-3.1.43-RELEASE] - 2025-06-20
+
+### Fixed
+- Hide "Start challenge" button when the user role is ADMIN (Taiga [#527], PR [#651])
```
Victorreca: Just to confirm — the changelog entry about hiding the 'Start challenge' button for ADMIN role (Taiga [#527], PR [#651]) seems related to a different branch. Should it be included in this PR, or was it added here by mistake?"

src/app/modules/challenge/components/challenge-form/challenge-form.component.css
```diff
@@ -331,3 +331,17 @@
...
   }
 }
+.link {
+  display: inline-flex;
+  align-items: center;
+  margin-bottom: 0.75rem; 
+}
+
+.breadcrumb {
+  display: block;
```
Victorreca, line 341: I think the display: block; style could be omitted — it doesn't seem to have any effect here.

src/app/modules/challenge/components/challenge-form/challenge-form.component.css
```diff
@@ -331,3 +331,17 @@
...
+  margin-bottom: 0.75rem;
+}
+
+.form-title {
```
Victorreca: The .form-title { margin-top: 0; } rule might be unnecessary — I didn’t notice any visible change with or without it. Maybe we could consider removing it?

src/app/modules/challenge/components/challenge-form/challenge-form.component.html
```diff
@@ -2,7 +2,7 @@
 <div>
     <!-- Header con título y botones -->
     <div class="header">
-      <div class="header-content">
+      <div class="header-content mt-4 mb-4">
```
Victorreca: You could replace mt-4 mb-4 with my-4 for a cleaner and more concise class name.

Victorreca (commented):
Great work overall! 🙌
I’ve left a few minor suggestions regarding styles that might be unnecessary or could be simplified (like display: block;, margin-top: 0;, or combining mt-4 mb-4 into my-4).
Feel free to keep them if there's a specific reason — just wanted to flag them in case they can be cleaned up.

src/app/modules/challenge/components/challenge-form/challenge-form.component.css
Silvitui: You're totally right. I checked and removing .form-title { margin-top: 0; } doesn't cause any visual regression. We originally added it to avoid default spacing from heading tags, but it seems unnecessary now. ✅ Removed!

src/app/modules/challenge/components/challenge-form/challenge-form.component.html
Silvitui: Thanks for the suggestion. I've updated the class to use my-4 instead of separate mt-4 and mb-4...it's definitely cleaner.

CHANGELOG.md
Silvitui: that changelog entry was from a different branch (Taiga #527). I accidentally included it during a rebase 😅 but I’ve fixed it since then. I hope everything looks good now .

ivilarop (approved):

PR #547, opened 2025-03-13
Feature/119 button to challengeform

GuillemBo:

PR #662, opened 2025-07-07
feat(challenge): allow user to save partial solution and continue later (#547)

Silvitui:
### 🧩 Context

This PR implements the functionality for users to save a **partial solution** while working on a challenge, and resume it later without losing progress. Defined in task #547 User story : #541.
### ⚠️ Note

The branch name is incorrect — it should be `feat/547-save-partial-solution` instead of `feat/541-save-partial-solution`.  

---

### ✅ What’s included

- 💾 **Partial solution saving**: Users can now write part of their solution and click "SAVE" to store their progress. The solution is saved in the database with status `'IN_PROGRESS'`.

- 🔁 **Challenge state detection**:
  - If a challenge is **not started**, the button **"Iniciar repte"** is shown.
  - If it was already started but not finalized (`IN_PROGRESS`), the user sees **"Continuar repte"**, along with **"Guardar"** and **"Finalitzar"** buttons.
  - If the challenge was **finalized** (`ENDED`), the user's solution is shown but cannot be edited.

- 🧼 **Reset editor content properly**:
  - If a user enters a challenge that has **no saved solution**, the editor now **starts empty**.
  - This prevents showing stale content from a previous challenge.

- 👤 **Role-based UI**:
  - Admin users do not see challenge action buttons (start/save/finish), only students (USER role) do.

- ⚙️ **Refactored logic** to ensure that solution state (`IN_PROGRESS`, `ENDED`) is correctly set based on backend data.

---

### 🧾 Technical note / Backend alignment required

To fully support this frontend feature, the backend should add support for a new solution status: `"NOT_STARTED"`.

Currently, the frontend handles this as a **UI-only state** to show the “Start challenge” button when no solution exists yet. However, to ensure full synchronization between frontend and backend, this state should ideally be recognized on the backend as well.

**Recommendation:**  
Extend the backend `UserSolutionStatus` enum (or equivalent logic) to include:

NOT_STARTED = 'NOT_STARTED'

### 🧪 How to test

1. Open any challenge as a **USER**, click **"Iniciar repte"**.
2. Type code like `console.log("test")`, click **"Guardar"**.
3. Navigate to a different challenge:
   - If it has no previous solution, the editor should be **empty**.
   - If it has a saved solution (`IN_PROGRESS`), it should load that content.
4. Navigate back to the original challenge → the previously saved code should appear.
5. Finalize a challenge and ensure the buttons disappear and the solution becomes read-only.
6. Log in as **ADMIN** and confirm that challenge action buttons are not displayed.

---
---

---

#### ✅ Solution already submitted
Here we have the view after the solution has been finalized. The challenge is marked as completed, and no editing is possible.

Solution submitted

---

#### 🔄 Continue challenge
This screen appears when the user has previously saved a partial solution. The "Continue Challenge" button allows the user to resume where they left off.

Continue challenge

---

#### 💾 Save or finalize
After clicking "Continue", the user sees the editor along with the "Save" and "Finish" buttons, giving them the choice to continue working or submit their final solution.

Save or finalize

#### `user-solution.interface.ts`
- ➕ Added `status` field to track the current state of the user’s solution (`IN_PROGRESS` or `ENDED`).
- 🎯 Enables partial saving and distinguishes between a draft and a finalized challenge.

#### `challenge-header.component.ts`
- 🧠 Detects challenge state dynamically: `NOT_STARTED`, `IN_PROGRESS`, `ENDED`.
- 🔄 Triggers reset of the editor when switching to a new challenge without saved text.
- 👤 Controls UI based on user role (ADMIN or USER) and challenge state.

#### `challenge-header.component.html`
- 👀 Shows/hides action buttons (`Start`, `Continue`, `Save`, `Finish`) based on the current solution state.
- 🔒 Prevents showing mentor’s solution until the user has finalized the challenge.

#### `editor-challenge.component.ts`
- 🧩 Receives solution text via `@Input()` and uses `ngOnChanges` to update content reactively.
- ✍️ Emits updated text to parent via `@Output()` every time the editor changes.
- 🔁 Resets textarea correctly when switching to a challenge with no saved solution.

#### `challenge-info.component.ts`
- 🔄 Manages the current state of the solution and toggles tab visibility.
- 🧠 Keeps logic centralized ).
- 👂 Listens to header and editor component events.

#### `challenge-info.component.html`
- 📌 Controls visibility of tabs and messages depending on solution state:
  - ❌ No tab shown if not started.
  - ✍️ Shows only user’s solution if `IN_PROGRESS`.
  - ✅ Shows both user and mentor solutions if `ENDED`.
  

  - #### `editor-challenge.component.ts`
- ❌ Removed localStorage auto-save on editor content changes.  
- 🔄 Now emits content changes via events and updates service only.

#### `modals/send-solution-modal/send-solution-modal.component.ts`
- 📤 Handles `END` status update on challenge finalization.
- 🧭 Ensures proper solution submission flow and modal behavior post-submit.

#### `modals/send-solution-modal/send-solution-modal.component.html`
- ✅ Confirms submission visually.
Removed localStorage auto-save from editor

### ✅ Why was this test removed?
We deleted the logic inside `onStartChallenge()` that stored data in `localStorage`, like:

```ts
localStorage.setItem('challengeStarted', JSON.stringify({ id: this.idChallenge, started: true }));
localStorage.setItem('currentChallengeId', this.idChallenge);
this.startChallenge.emit(true);
await this.router.navigate(['/ita-challenge/challenges', this.idChallenge, 'start']);

Now, we don’t need to use it...Keeping the localStorage logic caused problems, because we had two sources of truth:
localStorage
Backend 
This sometimes made the "Start Challenge" button appear twice or at the wrong time.

### CODE REVIEW

**Comment by oscarrep** [comment]
I have tested these changes and confirmed they are working as intended: as USER i can save progress on a challenge and come back to it later to finish it, and as ADMIN i can't see the challenge action buttons.

Great work overall! Let's hope you can pass the Sonar checks 🙏

**Review by oscarrep — APPROVED** [review]

File: et --hard ae3c640f
Comment (ivilarop) [inline et --hard ae3c640f:1]: what is this file?

**Comment by ivilarop** [comment]
> * NOT_STARTED

We don't have a NOT_STARTED status in the backend. What should we do?

image

image

File: src/app/models/user-solution.interface.ts [hunk 1]
```diff
@@ -3,6 +3,7 @@ export interface UserSolution {
   uuid_challenge: string
   uuid_language: string
   solution_text: string
+  status: 'IN_PROGRESS' | 'ENDED'
```
ivilarop: I see that in user-solution.interface.ts you've declared the status type as a union of literals 'IN_PROGRESS' | 'ENDED'.
Why not just reuse the SolutionStatus enum you just created?
This way we maintain a single source of truth for the statuses and avoid duplicate strings.

src/app/modules/challenge/components/challenge-header/challenge-header.component.html
```diff
@@ -49,8 +49,11 @@ <h3 class="ms-2">{{title | dynamicTranslate}}</h3>
 
   <div class="col-md-4 button d-none d-md-flex">
     <ng-container *ngIf="!solutionSent; else solutionSentMessage">
+       <button
+      *ngIf="solutionState === 'IN_PROGRESS' && !challengeStarted && userRole !== USER_ROLE.ADMIN"
```
ivilarop: In *ngIf, you're using 'IN_PROGRESS' as a literal string. Since you've already created the SolutionStatus enum, could you reuse it here as well? This way, we avoid typos and maintain consistency throughout the app.

Also, I see that in *ngIf you're mixing solutionState === 'IN_PROGRESS' with !challengeStarted.

Does it make sense for them to coexist in validation?

src/app/modules/challenge/components/challenge-header/challenge-header.component.html
```diff
@@ -49,8 +49,11 @@ <h3 class="ms-2">{{title | dynamicTranslate}}</h3>
...
+class="btn btn-sm btn-primary"(click)="onContinueChallenge()">{{ 'modules.challenge.header.buttonContinueChallenge' | translate }}</button>
   <button
-  *ngIf="!challengeStarted && userRole !== USER_ROLE.ADMIN" class="btn btn-sm btn-primary" (click)="onStartChallenge()" >
+  *ngIf="solutionState === 'NOT_STARTED' && userRole !== USER_ROLE.ADMIN" class="btn btn-sm btn-primary" (click)="onStartChallenge()" >
```
ivilarop: The literal 'NOT_STARTED' appears in the template, but this value doesn't exist in the SolutionStatus enum or in the API.

What should we do?

Silvitui:
We don't have a NOT_STARTED status in the backend. But in the frontend, we added "NOT_STARTED" as a local UI state for a specific reason:
It helps us detect when the user hasn't started the challenge yet, meaning there is no solution stored in the database for that challenge and user. Based on that, we can display the "Start challenge" button instead of "Continue challenge" or "Challenge completed".
So it's not part of the backend model. It's just a derived state used on the client side to improve the user experience. 
@ivilarop 
I’ve updated the description to clarify that the logic was refactored to rely only on the backend solution status (IN_PROGRESS, ENDED).

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -40,69 +41,84 @@ export class ChallengeHeaderComponent implements OnInit {
...
   @Input() isFavorite: boolean = false
   @Input() isBookmarked: boolean = false
-  @Input() languageId!: string
+  @Input() languageId: string = '';
   @Input() timesSolved: number = 0
   @Output() startChallenge = new EventEmitter<boolean>()
   @Output() favoritesUpdated = new EventEmitter<number>()
+  @Input() solutionText: string = '';
+  @Input() status: string = '';
+  @Input() solutionState: 'NOT_STARTED' | 'IN_PROGRESS' | 'ENDED' = 'NOT_STARTED';
```
ivilarop: Magic Strings, Reuse the existing SolutionStatus enum

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -40,69 +41,84 @@ export class ChallengeHeaderComponent implements OnInit {
...
-    this.authService.getUserRole().subscribe(role => {
-      this.userRole = role
-    })
+      if (solution) {
+        this.status = solution.status;
 
-    // Verifica si el reto ya ha comenzado
-    if (this.challengeStarted) {
-      this.activeId = ChallengeTab.SOLUTIONS
+        if (solution.status === 'IN_PROGRESS') {
```
ivilarop: This block can be refactored to gain readability, avoid duplication, and correctly use the SolutionStatus enum.

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -40,69 +41,84 @@ export class ChallengeHeaderComponent implements OnInit {
...
-        console.error('Error fetching user solutions:', err);
-      },
-    });
+  this.authService.getUserId().subscribe(userId => {
+    this.userId = userId;
 
-    this.authService.getUserId().subscribe(userId => {
-      this.userId = userId;
+    if (!userId) {
+      console.error("Could not get User ID");
```
ivilarop, line 77: What happens after console.error? Does the flow continue to run? 

Would the if else be necessary after the refactor?

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -40,69 +41,84 @@ export class ChallengeHeaderComponent implements OnInit {
...
-
-    if (savedChallenge.id === this.idChallenge && savedChallenge?.started === true) {
-      this.challengeStarted = true
-      this.activeId = ChallengeTab.SOLUTIONS // Mostrar botones de guardar y enviar solución
-    }
-  }
 
   async onStartChallenge (): Promise<void> {
     this.challengeStarted = true
+    this.solutionState = 'IN_PROGRESS';
```
ivilarop: enum?

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -124,6 +140,9 @@ export class ChallengeHeaderComponent implements OnInit {
     })
     modalRef.componentInstance.idChallenge = this.idChallenge;
     modalRef.componentInstance.userId = this.userId;
+    modalRef.componentInstance.status = 'ENDED';
```
ivilarop: Magic String. enum?

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -151,10 +170,28 @@ export class ChallengeHeaderComponent implements OnInit {
...
 
-  saveChallenge (): void {
-    console.log('Guardando reto...')
+saveChallenge(): void {
+  if (!this.idChallenge || !this.languageId || !this.solutionText || !this.userId) {
+    console.error(' Faltan datos para guardar la solución');
+    return;
   }
 
+  this.status = 'IN_PROGRESS';
```
ivilarop: Magic String

src/app/modules/challenge/components/challenge-header/challenge-header.component.spec.ts
```diff
@@ -40,12 +40,16 @@ describe('ChallengeHeaderComponent', () => {
       fetchUserSolution: jest.fn().mockReturnValue(
         of([
           {
+            uuid_user: "user1",
             uuid_challenge: "testChallengeId",
+            uuid_language: "testLang",
             solution_text: "some solution",
+            status: "ENDED",
```
ivilarop: magic String

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -151,10 +170,28 @@ export class ChallengeHeaderComponent implements OnInit {
     this.startChallenge.emit(true)
   }
 
-  saveChallenge (): void {
-    console.log('Guardando reto...')
+saveChallenge(): void {
+  if (!this.idChallenge || !this.languageId || !this.solutionText || !this.userId) {
+    console.error(' Faltan datos para guardar la solución');
```
ivilarop: text in english

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -151,10 +170,28 @@ export class ChallengeHeaderComponent implements OnInit {
...
+    this.languageId,
+    this.userId,
+    this.status,
+    this.solutionText
+  ).subscribe({
+    next: () => {
+      this.solutionState = 'IN_PROGRESS';
+    },
+    error: (err) => {
+      console.error(' Error al guardar solución', err);
```
ivilarop: text in english

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -151,10 +170,28 @@ export class ChallengeHeaderComponent implements OnInit {
...
+  this.status = 'IN_PROGRESS';
+  this.solutionService.submitSolution(
+    this.idChallenge,
+    this.languageId,
+    this.userId,
+    this.status,
+    this.solutionText
+  ).subscribe({
+    next: () => {
+      this.solutionState = 'IN_PROGRESS';
```
ivilarop: magic String

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -213,4 +250,28 @@ export class ChallengeHeaderComponent implements OnInit {
       })
     }
   }
+  onContinueChallenge(): void {
+  this.challengeStarted = true;
+  this.isEditorChallengeVisible = true;
+  this.status = 'IN_PROGRESS';
```
ivilarop: magic string

src/app/modules/challenge/components/challenge/challenge.component.ts
```diff
@@ -38,15 +40,27 @@ export class ChallengeComponent implements OnInit, OnDestroy {
...
   isEditorChallengeVisible = false
-  startChallenge: boolean = false
   challengeStarted: boolean = false
   favoriteChallenges: string[] = []
   bookmarkedChallenges: string[] = []
+  isChallengeStatementVisible = true;
+  isFavorite: boolean = false
+  userSolution: UserSolution | null = null;
+  solutionText: string = '';
+  status: string = 'IN_PROGRESS';
```
ivilarop: magic String

src/app/modules/challenge/components/challenge/challenge.component.ts
```diff
@@ -38,15 +40,27 @@ export class ChallengeComponent implements OnInit, OnDestroy {
...
+  status: string = 'IN_PROGRESS'; 
+  languageId: string = '';
 
   private readonly route = inject(ActivatedRoute)
   private readonly router = inject(Router)
   private readonly challengeService = inject(ChallengeService)
   private readonly _authService = inject(AuthService)
+  private readonly solutionService = inject(SolutionService)
+  private userId: string = ''
+  public solutionState: 'NOT_STARTED' | 'IN_PROGRESS' | 'ENDED' = 'NOT_STARTED'
```
ivilarop: Same case as above. Check it out.

src/app/modules/challenge/components/challenge/challenge.component.ts
```diff
@@ -55,33 +69,35 @@ export class ChallengeComponent implements OnInit, OnDestroy {
...
-          })
-          this.challengeService.getUserBookmarks(userId).subscribe({
-            next: (bookmarks: string[]) => {
-              this.bookmarkedChallenges = bookmarks
-            },
-            error: (err) => {
-              console.error('Error getting bookmarks:', err)
-            }
-          })
+   this._authService.getUserId().subscribe(userId => {
```
ivilarop: This snippet is not tabbed and is difficult to read. Also:

You are using literal strings ('IN_PROGRESS', 'ENDED', 'NOT_STARTED') that are already defined in the SolutionStatus enum.
Could you reuse it to maintain a single source of truth and avoid typing errors?

There are also several chained subscribe() methods without error handling or separation of responsibilities.
If we leave it this way for simplicity, it would be nice to at least:

Add an error to each subscribe or use catchError.

Type correctly with enums (SolutionStatus).

Add context to logs if included (console.error(...)).

src/app/modules/challenge/components/challenge/challenge.component.ts
```diff
@@ -92,11 +108,15 @@ export class ChallengeComponent implements OnInit, OnDestroy {
...
-  onStartChallenge (started: boolean): void {
-    this.challengeStarted = started
-    this.startChallenge = started
-    this.isEditorChallengeVisible = started
-  }
+onStartChallenge(started: boolean): void {
+  this.challengeStarted = started;
+  this.isEditorChallengeVisible = started;
+
+  this.status = started ? 'IN_PROGRESS' : 'NOT_STARTED';
```
ivilarop: magic string

src/app/modules/challenge/components/challenge/challenge.component.ts
```diff
@@ -126,6 +146,28 @@ export class ChallengeComponent implements OnInit, OnDestroy {
       this.popularity = this.challenge.popularity
       this.languages = this.challenge.languages
       this.timesSolved = this.challenge.timesSolved
+      this.languageId = this.languages[0]?.id_language ?? ''
     })
   }
+onChallengeStart(): void {
+  this.challengeStarted = true;
```
ivilarop: Are so many boolean flags really necessary?

src/app/modules/challenge/components/editor-challenge/editor-challenge.component.ts
```diff
@@ -81,7 +92,8 @@ export class editorChallengeComponent implements OnInit, OnChanges, OnDestroy {
             EditorView.updateListener.of((update: ViewUpdate) => {
               if (update.docChanged) {
                 const content = this.editor.state.doc.toString()
-                localStorage.setItem('editorContent', content)
```
ivilarop, line 84: Is this change documented in the PR description?

I don't see anything in it about changes to the localStorage

ivilarop (changes requested):

RonniedeTal:
I found some bugs in the challenge saving process, but it's working. Check it out.

-should be a continue button 
Captura de pantalla 2025-07-14 122721

-when you continue the challenge appears this problem at the console, is working, take a little look
Captura de pantalla 2025-07-14 123431

-when i start a challenge i have to press two times start challenge button to actually started and see the save button
1. 
Captura de pantalla 2025-07-14 124855
2. 
Captura de pantalla 2025-07-14 124903
3. 
Captura de pantalla 2025-07-14 124910

oscarrep (approved):
solutionStatus enum used correctly in all necessary places now!

src/app/modules/challenge/components/editor-challenge/editor-challenge.component.ts
Silvitui: I’ve already documented this change in the description. Let me know if you think it needs to be clearer or if you'd like me to add more detail.

et --hard ae3c640f
Silvitui: I’ve already updated the changelog to reflect it. Let me know if anything else is missing

src/app/models/user-solution.interface.ts
Silvitui: Yes..it's better to use the SolutionStatus enum. I've changed it now in the interface. Thanks for the suggestion

src/app/modules/challenge/components/challenge-header/challenge-header.component.html
Silvitui: Thanks! You're right. I’ll use the SolutionStatus enum instead of the string.
About the second part, I’ll check the logic to see if we really need both conditions.

src/app/modules/challenge/components/challenge-header/challenge-header.component.html
Silvitui: You're right ..'NOT_STARTED' is not part of the backend or the SolutionStatus enum.
We added it locally as part of the SolutionState type to handle the case when there's no solution yet.
It helps us manage the UI before the user starts a challenge, without needing extra values from the backend.

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
Silvitui: i’ve updated the solutionState to use the SolutionStatus enum and removed the magic strings.

src/app/modules/challenge/components/challenge/challenge.component.ts
Silvitui: The same for this. thank you

src/app/modules/challenge/components/challenge/challenge.component.ts
Silvitui: I've fixed the indentation to make it easier to read.
I’ve also replaced the string literals with SolutionStatus enum values to avoid typos.
For the observables, I’ve added proper error handlers and improved the console.error messages with more context.
Let me know if you see anything else to improve!

src/app/modules/challenge/components/challenge/challenge.component.ts
Silvitui: i’ve updated the code to use the SolutionStatus enum instead of the string literals. Thanks!

src/app/modules/challenge/components/challenge/challenge.component.ts
Silvitui: in this method I'm already using challengeStarted to control the state.
I’ll review if solutionState and other flags like isEditorChallengeVisible are really needed, or if we can simplify them. Thanks!

et --hard ae3c640f
oscarrep, line 1: remember to remove this unnecessary file

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -8,6 +8,9 @@ import { SolutionService } from 'src/app/services/solution.service'
 import { AuthService } from 'src/app/services/auth.service'
 import { ChallengeTab } from 'src/app/shared/enums/challenge-tab.enum'
 import { UserRole } from 'src/app/shared/enums/user-role.enum'
+import { SolutionStatus } from 'src/app/models/user-solution-status.enum'
+type SolutionState = SolutionStatus | 'NOT_STARTED';
```
oscarrep: should 'NOT_STARTED' go in the SolutionStatus enum?

src/app/modules/challenge/components/challenge-info/challenge-info.component.ts
```diff
@@ -68,6 +72,12 @@ implements OnInit {
 
   @Input() isEditorChallengeVisible: boolean = false
   @Input() startChallenge: boolean = false
+  @Output() solutionChanged = new EventEmitter<string>();
+  @Input() languageId: string = '';
+ @Input() initialSolutionText: string = '';
+@Output() solutionChange = new EventEmitter<string>();
```
oscarrep: these declarations are not properly aligned. keeping consistent indentation helps improve code readability.

src/app/modules/challenge/components/challenge/challenge.component.ts
```diff
@@ -10,6 +10,12 @@ import { type Example } from 'src/app/models/challenge-example.model'
 import { type Language } from 'src/app/models/language.model'
 import { ChallengeTab } from 'src/app/shared/enums/challenge-tab.enum'
 import { AuthService } from 'src/app/services/auth.service'
+import { SolutionService } from 'src/app/services/solution.service'
+import { UserSolution } from 'src/app/models/user-solution.interface'
+import { SolutionStatus } from 'src/app/models/user-solution-status.enum';
+type SolutionState = SolutionStatus | 'NOT_STARTED';
```
oscarrep: would including NOT_STARTED in the enum prevent the need to declare it here too?

src/app/modules/challenge/components/challenge-header/challenge-header.component.html
```diff
@@ -50,8 +50,11 @@ <h3 class="ms-2">{{title | dynamicTranslate}}</h3>
 
   <div class="col-md-4 button d-none d-md-flex">
     <ng-container *ngIf="!solutionSent; else solutionSentMessage">
+       <button
+      *ngIf="solutionState === SolutionStatus.IN_PROGRESS && !challengeStarted && userRole !== USER_ROLE.ADMIN"
+class="btn btn-sm btn-primary"(click)="onContinueChallenge()">Continuar repte</button>
   <button
-  *ngIf="!challengeStarted && userRole !== USER_ROLE.ADMIN" class="btn btn-sm btn-primary" (click)="onStartChallenge()" >
+  *ngIf="solutionState === NOT_STARTED && userRole !== USER_ROLE.ADMIN" class="btn btn-sm btn-primary" (click)="onStartChallenge()" >
```
oscarrep: could we use 'solutionState === SolutionStatus.NOT_STARTED'?

oscarrep (changes requested):

src/app/modules/challenge/components/challenge-info/challenge-info.component.ts
Silvitui: I've fixed! thank you

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
Silvitui: The backend doesn’t return a status like "NOT_STARTED", so we added it manually on the frontend.
We use it to know when a user has not started a challenge yet .This helps us show the "Start Challenge" button only when needed.

src/app/modules/challenge/components/challenge-header/challenge-header.component.html
Silvitui: we can’t use SolutionStatus.NOT_STARTED because is not part of the SolutionStatus enum.

src/app/modules/challenge/components/challenge/challenge.component.ts
Silvitui: we keep NOT_STARTED as a separate string in a union type
This way, the enum matches only real backend values

et --hard ae3c640f
Silvitui: I’ve already removed it, thanks!

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
oscarrep: okay! got it

oscarrep (approved):
Everything is good now

et --hard ae3c640f
ivilarop, line 1: This file is still here

src/app/modules/challenge/components/challenge-header/challenge-header.component.html
ivilarop: The conversations say one thing and the code another.

src/app/models/user-solution-status.enum.ts
```diff
@@ -1,3 +1,4 @@
 export enum SolutionStatus {
-    ENDED = "ENDED"
-  }
\ No newline at end of file
+  IN_PROGRESS = "IN_PROGRESS",
```
ivilarop, line 2: and the status NOT_STARTED ?

src/app/modules/challenge/components/challenge-header/challenge-header.component.html
ivilarop: About "As for the second part, I'll review the logic to see if we really need both conditions."

What conclusion did you reach?

src/app/modules/challenge/components/challenge-header/challenge-header.component.html
```diff
@@ -50,8 +50,11 @@ <h3 class="ms-2">{{title | dynamicTranslate}}</h3>
 
   <div class="col-md-4 button d-none d-md-flex">
     <ng-container *ngIf="!solutionSent; else solutionSentMessage">
+       <button
+      *ngIf="solutionState === SolutionStatus.IN_PROGRESS && !challengeStarted && userRole !== USER_ROLE.ADMIN"
```
ivilarop, line 54: About your comment: As for the second part, I'll review the logic to see if we really need both conditions.

What conclusion did you reach?

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
ivilarop: Should we include NOT_STARTED status in backend?

In the meantime, can we use enum instead of magic string?

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
Silvitui: Yes! I think  (!userId) return is better than using if/else here.
It's shorter and easier to read. we just stop the code early if there is no user ID.

src/app/modules/challenge/components/challenge-header/challenge-header.component.spec.ts
```diff
@@ -237,5 +231,158 @@ describe('ChallengeHeaderComponent', () => {
...
+
+  it('should navigate to challenges on cancel', () => {
+    component.onCancel();
+    expect(router.navigate).toHaveBeenCalledWith(['/ita-challenge/challenges']);
+  });
+  
+  it('should set solutionState to NOT_STARTED if no matching solution is found', () => {
+    solutionService.fetchUserSolution.mockReturnValue(of([]));
+    component.loadUserSolutionStatus();
+    expect(component.solutionState).toBe('NOT_STARTED');
```
ivilarop: enum

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -20,13 +23,17 @@ export class ChallengeHeaderComponent implements OnInit {
     private readonly modalService: NgbModal,
     private readonly translate: TranslateService,
     private readonly route: ActivatedRoute
+    
   ) {}
-
+  public SolutionStatus = SolutionStatus;
+  public NOT_STARTED = 'NOT_STARTED';
```
ivilarop: enum

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -40,81 +47,94 @@ export class ChallengeHeaderComponent implements OnInit {
...
+      const solution = userSolutions.find(
+        (sol) =>
+          sol.uuid_challenge === this.idChallenge
+      );
 
-    this.authService.getUserRole().subscribe(role => {
-      this.userRole = role
-    })
+      if (!solution) {
+        this.solutionState = 'NOT_STARTED';
```
ivilarop: enum

src/app/modules/challenge/components/challenge-header/challenge-header.component.ts
```diff
@@ -40,81 +47,94 @@ export class ChallengeHeaderComponent implements OnInit {
...
+          this.challengeStarted = true;
+          this.activeId = ChallengeTab.SOLUTIONS;
+          break;
 
-    if (savedChallenge.id === this.idChallenge && savedChallenge?.started === true) {
-      this.challengeStarted = true
-      this.activeId = ChallengeTab.SOLUTIONS // Mostrar botones de guardar y enviar solución
+        default:
+          console.warn(`Unhandled solution status: ${solution.status}`);
+          this.solutionState = 'NOT_STARTED';
```
ivilarop: enum

src/app/modules/challenge/components/challenge-header/challenge-header.component.html
Silvitui: I added the condition !challengeStarted because without it, the "Continue challenge" button is always visible.. even when the challenge has already started.
This condition hides the button when the challenge is running, and only shows it when the user hasn’t started it yet but has a saved draft.
In the future, we can refactor the logic to improve this

ivilarop (approved):

PR #677, opened 2025-07-24
Fix (challenge): Show editor when starting challenge from non-details tab (#632)

oscarrep:
Description:
This PR fixes an issue where the code editor was not shown when starting a challenge from a tab other than “Details”.

---

🛠 What was happening:
The `<app-editor-challenge>` component is located inside the “Details” tab, so it only renders when activeId === ChallengeTab.DETAILS.
When users clicked “Start Challenge” or “Continue Challenge” from another tab, the challenge state updated correctly but the UI remained on the current tab — leaving the editor hidden.
image

---

✅ Fix applied:
In challenge.component.ts, the method onStartChallenge() now explicitly sets activeId = ChallengeTab.DETAILS so that the editor becomes visible immediately when the challenge starts.
image

Scope of the change:

No HTML or layout changes were made. This is a minimal, non-breaking fix to improve user experience.
A future refactor is recommended to move the editor outside the tab system.

---

🧪 How to test:

1. Navigate to any challenge.
2. Switch to a different tab (e.g., “Resources”).
3. Click “Start Challenge” or “Continue Challenge”.
4. The UI should automatically return to the “Details” tab and show the editor.


RonniedeTal (approved):
is working

RonniedeTal (approved):
control version updated: done!!

ivilarop (approved):

PR #606, opened 2025-05-12
Feature/329 show favorite on login

martaag18:
This PR completes task 329, part of user story 314 – "As a mentor, I can give/remove likes and see my liked challenges."

**Functional changes:**

- When a user logs in, their list of liked challenges is fetched from the backend.

- On the challenge list page, the heart icon is automatically filled for challenges the user has already liked.

- On the challenge detail page, the heart icon in the header is also filled for liked challenges.

**How to Test It:**

1. Log in with a user that has liked some challenges.

2. Navigate to the challenge list view:

3. Confirm that liked challenges show a filled heart icon.

4. Click on any liked challenge to go to the detail view:

5. The header should also display a filled heart icon.

image
image

PR #679, opened 2025-07-28
Feature(register-users): Modal register users backend connection (#606)

oscarrep:
This PR finalizes task **606**, part of user story **582**

### 🧩 Context
This task connects the refgister users modal logic to the project's backend

### ✅ What was done

- Removed the mocked logic `register-users.service`

#### ⚠Disclaimer

Had to remove ``CREATE_USER`` from environment variables and hardcode it on ``register-users.service.ts`` since adding it there made duplicated code go over 3% and SonarQube did not allow it. For the moment this was the only viable way to go about it since the proper way to solve it would have been refactoring all shared variables from `environment.ts` and `environment.prod.ts` into a shared environment file.

---

### 🧪 How to test
1. Log in as a mentor (ADMIN role)
2. Click on `Add students`
image

3. Type a GitHub username and click `Add` — it should appear in a list
image

4. After adding one or multiple usernames, click `Confirm` — If the usernames are not already registered it should show `Users registered successfully`
image

5. If the user already exists in the database or any other problem occured, the app will show an according message - In a later update, it would be ideal to change it so it shows that one or more of the users added already exist in de database, ideally mentioning which users are triggering the error.
image

---

### 📁 Files changed
- `environment.ts`
- `environment.prod.ts`
- `register-users.service.ts`
- `register-users.service.spec.ts`
- `register-users-modal.component.html`
- `register-users-moda.component.ts`

---

### ✅ Auto self-check:
- [x] I tested my changes locally and verified they work as expected
- [x] The PR has a clear and focused purpose (only one feature, fix or refactor)
- [x] Title, description, and changelog (if needed) are filled in correctly
- [x] There are no leftover merge conflicts or unrelated changes from previous branches
- [x] All automated checks (like SonarCloud) have passed
- [x] I responded to all previous review comments and only resolved those I truly addressed
- [x] I ran SonarLint locally on the modified files and confirmed there are no warnings or errors


RonniedeTal (approved):
seems correct !!!

src/app/services/register-users.service.ts
```diff
@@ -8,13 +8,7 @@ export class RegisterUsersService {
   constructor(private readonly http: HttpClient) { }
 
   registerUser(username: string): Observable<any> {
-    return this.http.post(`${environment.BACKEND_ITA_CHALLENGE_BASE_URL}/${environment.CREATE_USER}`, { username });
+    return this.http.post(`${environment.BACKEND_ITA_CHALLENGE_BASE_URL}/admin/users/create`, { username });
```
oscarrep, line 11: I'm aware that /admin/users/create should be included in the environment files, but adding them there triggered sonarqube duplication and this was the only way i found that wouldn't trigger it. other than refactoring both files into a shared one.

ivilarop: OK! Thanks for commenting. Can you create a technical debt so we can address it in future sprints? If you could share the problem and solution, that would be great.

ivilarop:
@oscarrep  Are the test cases documented if the user already exists? Perhaps it would be helpful to add screenshots to demonstrate this case.

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [ita-challenges-frontend-3.2.4-RELEASE] - 2025-07-28
```
ivilarop: Should it be a minor change? Please justify it.

ivilarop (changes requested):

src/app/services/register-users.service.ts
oscarrep: Good idea, I created it

CHANGELOG.md
oscarrep: My reasoning for treating it as a patch was that the user wouldn't really see a change in the behavior of the app.
Since this change actually connects front and back, it could also be argued as a minor change.

oscarrep:
> @oscarrep Are the test cases documented if the user already exists? Perhaps it would be helpful to add screenshots to demonstrate this case.

RIght now it only mentions that something went wrong, but it would be ideal in a later update. Screenshot added

CHANGELOG.md
oscarrep: Minor change is a better fit, changed to 3.3.0

ivilarop: Ok thks!

ivilarop:
> > @oscarrep Are the test cases documented if the user already exists? Perhaps it would be helpful to add screenshots to demonstrate this case.
> 
> RIght now it only mentions that something went wrong, but it would be ideal in a later update. Screenshot added

Ok thanks, we've added a technical debt to change the return message to be more explicit: https://tree.taiga.io/project/luisvi-ita-challenges/issue/639

ivilarop (approved):

PR #648, opened 2025-06-13
fix(solution tab in challenge-info): prevent execution without userId, improve error handling, and minor UI fixes (#512)

basantades:
This PR completes **task 512**

The challenge-info component previously executed functions before userId was available, causing issues when refreshing the page. Additionally, the service had uncontrolled error loops when fetching user solutions, and two UI inconsistencies were present.

##  Changes:
- Prevented execution of functions before userId is available, ensuring correct data loading.
- Improved error handling in the service, preventing infinite loops and unnecessary calls.
- Corrected the admin solution display, ensuring it appears in a single column.
- Removed erroneous text ("Escriu el teu codi aquí") from finalized user solutions. [Technical Debt](https://tree.taiga.io/project/luisvi-ita-challenges/issue/456)

image

### Modified Files:

- src/app/services/solution.service.ts
- src/app/services/solution.service.spec.ts
- src/app/modules/challenge/components/challenge-info/challenge-info.component.ts
- src/app/modules/challenge/components/challenge-info/challenge-info.component.html
- src/app/shared/components/solution/solution.component.html
- CHANGELOG.md

### 👩‍🎓 As a Student:
1. **Complete a challenge** and verify that it redirects you to the challenge's solution tab.  
   - Ensure both **your solution** and the **mentor's solution** are displayed correctly.  
   - Check the console to confirm that **no error loops** occurred when loading solutions.  
2. Verify that the **student solution** does **not display** the text `"Escriu el teu codi aquí"`.  
3. **Refresh the browser** and re-enter the solutions tab.  
   - Confirm that the information loads correctly.

### 🛠️ As an Admin:
1. **Navigate to the solutions tab** for a challenge.  
2. **Ensure that the solution is displayed correctly in a single column**.

## ✅ PR Author Self-check  

- [x] I tested my changes locally and verified they work as expected.  
- [x] The PR has a clear and focused purpose (only one feature, fix or refactor).  
- [x] Title, description, and changelog (if needed) are filled in correctly.  
- [x] There are no leftover merge conflicts or unrelated changes from previous branches.  
- [x] All automated checks (like SonarCloud) have passed.  
- [x] I responded to all previous review comments and only resolved those I truly addressed.

PR #684, opened 2025-09-10
chore(contributors): add-Leonid-Mozheliuk to contributors.md(#648)

Leoyalta:
Added Leonid Mozheliuk to the contributor’s list as part of the onboarding process.


VictorGonPe (commented):
Hi Leo, thank you for the contribution! The implementation looks good 👍  
Just a quick note: the task number used is not the one from Taiga.  
It would be better to update it so we can keep consistency with the project’s task tracking.

VictorGonPe (approved):
Perfect! The changes were made correctly.

PR #709, opened 2025-11-07
fix(challenge-form): fix edit challenge tags (#845)

FlavioKde:
### 🎯 Summary

This PR corresponds to task [#845](https://tree.taiga.io/project/luisvi-ita-challenges/task/845). It is the first task from HU [#816](https://tree.taiga.io/project/luisvi-ita-challenges/us/816), the main goal of this HU is to fix an issue regarding selected tags during challenge edition.

When editing a programming challenge, previously selected tags do not appear active in the form. 
Because of that, if the mentor re-selects the same tag (which is already stored), the frontend sends a duplicated payload such as:

Tags during edition mode:
image

Tags selected during edition mode:
image

`"tags": ["0c8ca47c-d915-486e-8ae9-803483c8ed17", "0c8ca47c-d915-486e-8ae9-803483c8ed17"]`

This causes a 400 Bad Request from the backend ("tag UUID duplicated: ...") and the user sees no message in the UI.

image

--------------------

### 🧠 The fix ensures:

- Stored tags appear preselected (active) when editing a challenge.

- The form sends a deduplicated tag array matching the UI selection.
 
- Validation messages (e.g. “Tags required”) display properly.
 
- The user never sees a silent failure.

--------------------

### 🧩 Key Changes

- Updated `loadTags()` to set `tagsControl` verifying tags initialization

- Updated `onSubmit()` to set `save logic` correctly bringing the persisted tags

- Added unit tests covering both scenarios 

--------------------
### 🔄 Modified files

- /Users/pablofedericogomez/ita-challenges-frontend/src/app/modules/challenge/components/challenge-form/challenge-form.component.ts
 
- /Users/pablofedericogomez/ita-challenges-frontend/src/app/modules/challenge/components/challenge-form/challenge-form.component.spec.ts
 
- /Users/pablofedericogomez/ita-challenges-frontend/package.json

--------------------

### 🔄 Changelog

Changelog updated

--------------------

### 🧪 Final result

- Open the edit form for a challenge with already selected tags during challenge creation→ form loads correctly.

- Open the edit form for a challenge with tags, you will be able to adding or removing tags, and submitting the challenge edited→ endpoint PUT response with a 200 OK.

Edition mode after frontend changes:
image

image

Adding a tag in edition mode after frontend changes:
image

API response  after adding changes:
image

Outcome:
image

Removing two tags and adding another one in edition mode after the frontend changes:
image

Outcome:
image

- Verify that no console errors appear.

--------------------

###  ✅  PR Author Self-Check

- [x] Tested the flow locally (logged-in vs logged-out).
- [x] Unit tests pass for the new logic.
- [x] No unrelated changes or merge conflicts.


eduzavarce:
From the screenshots, I cannot say whats the original state and what is the outcome.

eduzavarce:
From the screenshots, i can assume you can add tags, but i do not see what happens if you revove them.

eduzavarce (commented):
@FlavioKde @Pablo2203 

Everything seems in order as discussed yesterday. 

only one doubt:
image

Why isnt this checkbox checked? are there conflicts?

Pablo2203:
@eduzavarce Hello Edu! Thank you for your insights. We agreed with Flavio about those changes you suggested. Now It is everything updated.

- [x] From the screenshots, I cannot say whats the original state and what is the outcome.
**Now there are epigraphs detailing the process states.**

- [x] From the screenshots, i can assume you can add tags, but i do not see what happens if you revove them.
**Examples added showing the tag remotion.**

- [x] Why isnt this checkbox checked? are there conflicts?
**Checkbox checked!**

eduzavarce (approved):
Good job, seems everything is in order now!

src/app/modules/challenge/components/challenge-form/challenge-form.component.ts
```diff
@@ -249,16 +253,19 @@ loadChallengeForEditing(): void {
...
-      },
-      error: (error) => {
-        this.currentTags = []
-        this.tagsControl.setValue([]);
-        throw error;
-      }
-    })
-  }
+        
+      this.tagsControl.setValue(this.selectedTags);
```
ivilarop, line 259: We could refactor this block to remove duplication and improve error handling:

- Simplify the if/else by computing the array first and calling setValue only once.

src/app/modules/challenge/components/challenge-form/challenge-form.component.ts
```diff
@@ -249,16 +253,19 @@ loadChallengeForEditing(): void {
...
+          } else {
+            this.tagsControl.setValue([]);
+          }
+        },
+        error: (error) => {
```
ivilarop, line 264: - We shouldn’t use throw inside the subscribe error callback.
Since the component already resets its state and gives user feedback, there’s no need to raise an exception here.
Instead, it’s cleaner to handle this earlier in the stream (using catchError) or just log the issue.
This avoids unhandled exceptions and makes the observable flow easier to maintain.

src/app/modules/challenge/components/challenge-form/challenge-form.component.ts
```diff
@@ -93,6 +93,10 @@ export class ChallengeFormComponent implements AfterViewInit, OnDestroy, OnInit
     translate.addLangs(['en', 'es', 'ca'])
   }
   ngOnInit(): void {
+  this.tagsControl.valueChanges.subscribe(value => {
+    this.selectedTags = value || [];
```
ivilarop, line 99: We can simplify the state: keep currentTags for the available/tag list, and use tagsControl.value as the single source of truth for the user selection.
selectedTags is redundant (it mirrors the control via valueChanges).

**Benefits**: fewer sources of truth, less sync code, fewer edge cases.

**Action items**:

1. Remove selectedTags and read the selection from tagsControl.value wherever needed (filter, submit, UI).
2. When loading tags, filter the control’s current value against currentTags and set the filtered value once.
3. Type currentTags como Tag[] (evitar any[]).
4. Maintain consistency in errors: if tag loading fails, leave currentTags = [] and tagsControl empty/invalid.

ivilarop (changes requested):

src/app/modules/challenge/components/challenge-form/challenge-form.component.ts
FlavioKde: Hello Ivan

What we did was adapt us to the code that it was already written. The synchronization was product of the two tag states pre existing, the only way to do not have residual tags, was the synchronization. Your comment is correct, it should be only one source of truth, but this is the way we found the code

We only fix the water leak from the pipe. If you want we can leave it as thecnical debt for future sprints

Pablo2203: @ivilarop 

Hello Ivan,

We have to add some more regarding this change you propose. The preexistent tests would fail if we did a change like that, it would be compromise this part as well. We do not know the reason and the scope of this two states inside the project.

src/app/modules/challenge/components/challenge-form/challenge-form.component.ts
FlavioKde: The throw error is aligned with the rest of the file; it is handled in the same way, so it preserves the semantics of the file.
We can use console.error to align with your solution; it logs the error to the console without breaking the flow.
The component recovers from the error by resetting the state.
If I want to provide feedback to the user, I should modify the HTML,
which would be outside the scope of our responsibility in this user story.

In the onSubmit() method, I use commonModalService.errorPostingChallengeModal(),
but that method is designed to show errors related to posting challenges,
so the modal’s specific message would not be appropriate for a tag-loading error.
If we want to use that approach, we should create a generic method or a more specific one.

We believe that using console.error would be appropriate, since the error would end here if we do not want to modify other parts of the code or use other options to handle it.
We can place the console.error instead of what is currently there.


```
error: (error) => {
          this.currentTags = [];
          this.tagsControl.setValue([]);
          //throw error;
          console.error('Error loading tags:', error);
        }
      });
```

src/app/modules/challenge/components/challenge-form/challenge-form.component.ts
FlavioKde: You're absolutely right about the duplication. Currently we have two synchronized states (tagsControl and selectedTags) due to existing technical debt. I kept the synchronization to fix the HU without introducing a larger refactor.

src/app/modules/challenge/components/challenge-form/challenge-form.component.ts
ivilarop: Thanks both for the clarification — understood 👍
The current approach works as a quick fix and keeps consistency with the existing codebase.
I fully agree we shouldn’t risk breaking pre-existing tests or UI flows in this PR.
Let’s document this as technical debt: in a future sprint we should refactor to have a single source of truth for tags (tagsControl.value) and remove the sync logic between selectedTags and currentTags.

✅ No changes required now — just make sure we mention this in the next tech-debt list or in the upcoming refactor task.

Please note that in the PR description.

src/app/modules/challenge/components/challenge-form/challenge-form.component.ts
ivilarop: For this PR, let’s avoid throw inside the subscribe. We’re already resetting state and giving user feedback, so a minimal log is enough here.

src/app/modules/challenge/components/challenge-form/challenge-form.component.ts
ivilarop: We'll leave that as a technical debt for a future task: "Unify tag selection: single calculation + single setValue (preserve errors/touched/dirty and synchronize selectedTags)." Not blocking here. 

It is included in the previous technical debt

ivilarop (changes requested):

src/app/modules/challenge/components/challenge-form/challenge-form.component.ts
Pablo2203: @ivilarop Perfect! Let´s do that. Please find the technical debt in this link: https://tree.taiga.io/project/luisvi-ita-challenges/issue/859

src/app/modules/challenge/components/challenge-form/challenge-form.component.ts
Pablo2203: @ivilarop @FlavioKde console.error implemented!

src/app/modules/challenge/components/challenge-form/challenge-form.component.ts
Pablo2203: @FlavioKde @ivilarop Hello again! Just in case as we agreed above, I left the link with the technical debt here:
https://tree.taiga.io/project/luisvi-ita-challenges/issue/859

Cheers!

ivilarop (approved):

PR #725, opened 2025-12-10
fix(challenge-info): fix user solutions loading to display correctly per challenge (#910)

Ilmira83:
**Summary**

This PR refactors how the frontend loads and manages user solutions to ensure strict per-challenge isolation.
It fixes an issue where partial or draft solutions unintentionally appeared in other challenges when navigating between them.

The main goal of this task was to audit the current frontend user solution handling eliminate cross-challenge state leakage and make the app rely exclusively on backend data and ensure that every challenge initializes its solution independently.

**Main Updates**
 - Challenge-info.components.ts - delete solutionSent$ subscription so now ngOnInit trigger a fresh API load loadUserSolutionData().


FlavioKde (approved):
Good job Ilmira!!

aperez1haad (approved):
good job

ivilarop:
@Ilmira83 

I've been testing locally with your branch and I see that solutions from one challenge are no longer being carried over to the rest, but when you publish a solution, it doesn't show both solutions and you have to refresh. The label with the solution's status isn't showing either. Is this normal behavior?

image

image

after refresh:

image

ivilarop:
Version? Changelog?

Ilmira83:
@ivilarop Ivan, I'v updated the RELEASE version - 3.14.1. as PATCH.

Ilmira83:
@ivilarop Regarding the UI behavior after clicking ACCEPT button - yes, at the moment it is like this. The lable was not shown for this button - it works for Show solution btn only. For both of my sprints the logic of ACCEPT btn was Out of Scope - so I believe I need to create an Issue for this. Also, the appearance of User's solution after submition is not handled clear enough in ACCEPT btn logic. So this is also the issue that should be resolved in order to handle properly the BE responses and use them in UI.

src/app/modules/challenge/components/challenge-info/challenge-info.component.ts
```diff
@@ -91,22 +89,14 @@ implements OnInit, OnDestroy {
...
-    });
-
     this.solutionService.activeIdSubject.next(ChallengeTab.DETAILS)
 
     this.solutionSent = this.solutions.includes(this.idChallenge)
     this.solutionService.solutionSent$.subscribe((sent) => {
       this.solutionSent = sent;
       if (sent) {
         this.loadSolutions(this.idChallenge, this.languages[0].id_language);
+        void this.loadUserSolutionData()
```
Ilmira83, line 99: This API call load the user solution after Submit.

src/app/modules/challenge/components/challenge-info/challenge-info.component.ts
```diff
@@ -58,7 +56,7 @@ implements OnInit, OnDestroy {
   private readonly authService = inject(AuthService)
   private readonly cdr = inject(ChangeDetectorRef)
   private readonly starterService = inject(StarterService) 
-   private readonly challengeService=inject (ChallengeService)
+  private readonly challengeService = inject(ChallengeService)
```
Ilmira83, line 59: Just added spaces between the words.

ivilarop: Okay, as we know, we wouldn't apply these changes because it makes reviews more difficult. In this case, it's a small change, and we see that you're aware of it. We'll accept it. We'll close the conversation. Thank you.

src/app/modules/challenge/components/challenge-info/challenge-info.component.ts
ivilarop: ok thsk! understood

ivilarop (approved):

ivilarop (approved):

PR #726, opened 2025-12-10
fix(core): resolve_loop_challenge(#921)

FlavioKde:
# **Problem solved:**
- Infinite loop (ERR_INSUFFICIENT_RESOURCES) when sending solutions
- Error caused by recursive sending in solutionSentSubject

---

## **Changes made:**
1. In `solution.service.ts`: Commented out the line `solutionSentSubject.next(value)` that was causing the loop
2. In `send-solution-modal.component.ts`: Use `SolutionAction.COMPLETED` (now working after backend fix)
3. Cleaned up unused imports

---

## **Trade-off accepted:**
- To eliminate the critical loop, automatic sending was disabled
- Consequence: The UI requires a manual refresh to see updates

---

## **Verification:**
- ✅ No more infinite loop
- ✅ Solutions are being sent correctly (200 OK)
- ✅ Bookmarks work
- ✅ Challenge is marked as completed (after refresh)

---

## **Technical-debt:**

[[TECHNICAL-DEBT-issue#928]](https://tree.taiga.io/project/luisvi-ita-challenges/issue/928) Fix broken menu links (componentToBuild)
- Problem: 4 links point to a non-existent route
- Solution: Change to valid routes or implement the functionality
- Priority: Low

[[TECHNICAL-DEBT-issue#936]](https://tree.taiga.io/project/luisvi-ita-challenges/issue/936) Delay in dual solution display
- Problem: A small delay was introduced as a workaround to avoid inconsistencies when displaying both solutions (user + itacademy). This is because the current state depends on a boolean that must be manually reset.
- Solution: Replace the boolean with an event in solution.service.ts
- Priority: Middle

[[TECHNICAL-DEBT-issue#937]](https://tree.taiga.io/project/luisvi-ita-challenges/issue/937) SubmitSolution does not broadcast 'submitted' on success
- Problem: Currently, submitSolution does not emit any events, which creates an inconsistency with the service requirements. A minimal test was included to reflect the expectation, but the implementation requires refactoring to emit "submitted" exactly once upon success.
- Solution: Refactor submitted in solution.service
- Priority: Middle


aperez1haad (changes requested):
You've a lot of lines and comments on the code that isn't running, is it a best practice?
you don't change the changeload too but im not sure if is needed.

aperez1haad (approved):

.gitignore
ivilarop, line 1: ??

src/app/modules/challenge/components/challenge-info/challenge-info.component.ts
```diff
@@ -100,17 +101,37 @@ implements OnInit, OnDestroy {
...
+
       this.solutionSent = sent;
+
       if (sent) {
         this.loadSolutions(this.idChallenge, this.languages[0].id_language);
       }
       this.cdr.detectChanges();
     });
 
+    this.solutionService.solutionSent$.pipe(
```
ivilarop, line 97: In challenge-info.component.ts, there are now two subscriptions to the same observable, solutionSent$ (one direct and one using distinctUntilChanged). In RxJS, each subscriber is an independent listener; when solutionSent$ emits, both execute, duplicating state allocation, data reloading, and detectChanges. This increases unnecessary work and can reopen the loop we were trying to avoid.

The difference between the two is that the first reacts to all emissions; the second only when the value changes. However, because they coexist, the first causes everything to happen twice, negating the filtering effect of the second.

Required action: keep only one subscription (the filtered one) and remove the other. Also, prevent solutionText$ from setting solutionSent = true, because it introduces a second source of truth for the same state. Result: a single reload per send, without duplicate effects or loops.

src/app/services/solution.service.ts
```diff
@@ -28,14 +28,21 @@ export class SolutionService {
   private readonly challengeCompletedSubject = new Subject<string>()
   challengeCompleted$ = this.challengeCompletedSubject.asObservable()
 
+  private isUpdating = false;
```
ivilarop, line 31: -----------------------------------------------------------
What to change NOW (in this PR) — minimal bug fix
---------------------------------------------------------------

Goal: Ensure the UI updates only once per submission and avoid duplicate sources.

1.-Emit "submitted" only from the service and only once per successful submission.

Move the emitter inside the submitSolution(...) when the PUT request returns OK.

If it fails, don't emit (and log/report the error).

2.-Don't emit "submitted" from other flows.

solutionText$ (and any other) should not be marked as "submitted".

3.-(Suggested and trivial) Expose solutionSent$ with a duplicate filter.

Apply distinctUntilChanged to the exposed observable (not on each component).

With this, even though we continue using the boolean, we guarantee a single automatic update per submission without loops. No further restructuring is needed in this PR.

---------------------------------------------------------------
What remains as TECHNICAL DEBT (TD-#932) — refactor
----------------------------------------------------------------
Goal: clean and re-entry-proof design.

1.-Change from boolean to event.

Treat “solution submitted” as an alert (event) that is only emitted when the submit is successful.

2.-Remove unnecessary code and confusing names.

Remove `updateSolutionSentState(...)`, `sendSolution(...)`, and any flags/manuals.

Make the event read-only and rename methods that are easily confused with observables (e.g., the text method).

3.-Ticket acceptance criteria.

1 successful submission ⇒ 1 event / 1 reload.

On error ⇒ 0 events, error message displayed.

Consecutive submissions ⇒ independent events, no duplicates.

-----------------------
One-line summary
--------------------

Now (PR): Issue only once on success from submitSolution, do not issue from other flows, and (if possible) filter duplicates when exposing solutionSent$.

Then (TD): Migrate to event-based reporting, clean up methods/flags, and set up behavior AC.

src/app/shared/components/challenge-card/challenge-card.component.ts
```diff
@@ -73,32 +73,39 @@ export class ChallengeCardComponent implements OnInit {
...
-      this.challengeService.addBookmark(this.id).subscribe({
-        next: response => {
-          this.isBookmarked = response.bookmarked
-        },
-        error: error => {
-          console.error('Error adding bookmark:', error)
-        }
-      })
-    }
+  event.stopPropagation()
```
ivilarop: Why this changes? Is this within the scope of the HU? I don't see any explanation in the PR description.

src/app/services/solution.service.spec.ts
ivilarop, line 1: To align the tests with the requirements for the component (a single subscription) and the service (emitting "submitted" only upon successful submission, without other sources), make these minimal changes:

What to remove

❌ should update solution sent state (use updateSolutionSentState)

❌ should send solution and update state (use sendSolution)

These methods should no longer be used in this pull request. Keeping the tests would force you to retain redundant code.

What to add/modify
----------------------------------------

Emit exactly once upon successful submission
Do not emit if the submission fails

ivilarop (changes requested):

DiegoBalaguer (approved):
Good research work on the problem you have managed to solve.

.gitignore
FlavioKde: solved, removed

src/app/modules/challenge/components/challenge-info/challenge-info.component.ts
FlavioKde: solved, removed the second suscription

src/app/services/solution.service.ts
FlavioKde: resolved, deleted the PR TECHNICAL DEBT (TD-#932)

What to change NOW (in this PR) — minimal bug fix--> changed now

src/app/shared/components/challenge-card/challenge-card.component.ts
FlavioKde: Resolved

src/app/services/solution.service.spec.ts
FlavioKde: Ivan, the redundant tests have been removed.

only implement one of the proposed tests, because there is an inconsistency in

submitSolution(uuid_challenge: string, uuid_language: string, uuid_user: string, action: string, solution_text: string): Observable<any> { 
const body = {--------------------------
}

It does not emit anything, as it is currently implemented and the test I added fails, it only makes an HTTP request

it('should emit "submitted" exactly once on successful submission', (done) => { 
const mockResponse = { success: true, message: 'Solution submitted successfully' }; 
let emissionCount = 0; 

service.sendSolutionText$.subscribe((submitted) => { 
if (submitted) { 
emissionCount++; 
} 
}); 

service.submitSolution(uuid_challenge, uuid_language, uuid_user, action, solution_text).subscribe({ 
next: (response) => { 
expect(response.success).toBe(true); 
expect(emissionCount).toBe(1); // only one broadcast 
done(); 
} 
}); 

const req = httpMock.expectOne(`${environment.BACKEND_ITA_CHALLENGE_BASE_URL}${environment.BACKEND_ITA_CHALLENGE_USER_SOLUTION}`); 
req.flush(mockResponse);
});

I left technical debt

.gitignore
ivilarop: ok thks!

ivilarop (approved):

src/app/modules/challenge/components/challenge-info/challenge-info.component.ts
ivilarop: ok thks!

src/app/services/solution.service.ts
ivilarop: ok thks!

src/app/shared/components/challenge-card/challenge-card.component.ts
ivilarop: ok thks!

src/app/services/solution.service.spec.ts
ivilarop: ok thks

PR #732, opened 2026-01-26
fix(challenge): synchronize challenges status logic Show solution and Finalize methods (#35)

Ilmira83:
This PR introduces the solutions of the task # 30 
https://github.com/IT-Academy-BCN/ita-challenges/issues/30

Sub-task # 35 
https://github.com/IT-Academy-BCN/ita-challenges/issues/35

Includes following changes:
- Added translation mapping for the SUBMITTED_INCOMPLETE status in i18n json files so now it is correctly set and displayed in both Challenge Details and Challenges List (Show solution btn).

- Added loadUserSolutionStatus() call in onSolutionAccepted() method (challenge-header.component.ts) in order to display correctly  the SUBMITTED_COMPLETE status after the challenge was Finished.

- Created new modal ShowSolutionModal that handles the logic behind the “Show solution” button by informing the users and allowing them to confirm or cancel the action.  

- The necessary translations for the ShowSolutionModal were included in i18n json files.

- Unit tests were accordingly updated and added.


Arnau-66 (approved):
Verified. The challenge status is now shown correctly after finishing, and the new confirmation modal clearly informs users that viewing the solution disables further edits. The behavior matches what was discussed in the original ticket and works as expected.

vaniaferreresteban (commented):
Hi! 

The code narrowly adress the problems and it solves them right, however i feel the scope is a bit unclear because the spected behavior of the state in progress is to not show the official solution ( and this is out of scope) and we have some duplicated checks because of that reason.

I've been working to align the logic between our PRs, and here are my conclusions:
### ❌ Request Changes
**1. Functional Regression**
The current conditional in [loadUserSolutionData](cci:1://file:///a:/Documents/Dev/ITAcademy/angular-java/ita-challenges-frontend/src/app/modules/challenge/components/challenge-info/challenge-info.component.ts:139:2-158:3) (based only on `match !== undefined`) is causing the official answer to be revealed prematurely, it's not the expected behavior to the backend state.
I understand that `solutionSent` returning `true` is technically semantically correct (the drafted solution *is* sent), but using it directly to control visibility creates this bug.
**2. Proposed Solution**
We can fix the behavior by adding the status check (using imported `SolutionStatus`) at line challanges-info.component.ts 152 :
```typescript
if (match !== undefined && match !== null && 
   (match.status === SolutionStatus.ENDED || match.status === SolutionStatus.SHOW_SOLUTION))
```

This fixes the bug immediately and avoids duplicade code with my PR making mine unnecessary. Although it makes solutionSent semantically confusing (returning false even if a draft was sent).

3. Alignment & Tech Debt

Pros: Using the SolutionStatus enum aligns with ChallengeHeader and brings us closer to a standardized model.
Cons: This logic still duplicates the fetch found in ChallengeHeader, creating "duplicated sources of truth" (Tech Debt) as i've also recieved by Ivan in my PR. 

Conclusion: I would add the Status check here to fix the visibility bug. This invalidates the need for my separate implementation in feature/44 since this code will cover it without any extra variables, and we can address the centralization/tech-debt in a future refactor.

vaniaferreresteban (approved):

vaniaferreresteban:
As we discussed at discord; we're going to advance and continue with the centralize refactor in a future, so let's just advance for now

ivilarop (approved):

PR #741, opened 2026-02-04
feat(frontend): 1.Migrate solution submission and retrieval to Challenge (#108)

anaberod:
## 🧩 Context

This PR migrates all solution-related frontend calls from the **User** microservice to the **Challenge** microservice, aligning the frontend with the current domain ownership and backend migration strategy defined in issue #95.

The frontend now interacts exclusively with the `SubmissionController` in Challenge for both **submission actions** and **user solution retrieval**.

---

## 🔧 What’s included

- 🔁 **Solution submission migrated**  
  SAVE / SUBMIT / GIVE_UP actions now call Challenge submission endpoints.

- 📥 **User solution retrieval migrated**  
  User submissions are fetched via Challenge endpoints instead of User.

- 🧹 **Legacy User endpoints removed from frontend**  
  No calls remain to:
  - `/user/solution`
  - `/user/users/{userId}/solutions`

- ⚙️ **Environment configuration updated**  
  Frontend environment variables now point to Challenge submission paths.

---

## 🧪 Evidence (Chrome DevTools – Network)

The migration was verified using **Chrome DevTools → Network → Fetch/XHR**.

Observed request example:
image

PR #771, opened 2026-02-21
fix(challenge) route edited in environment to meet real endpoint 

### LINKED ISSUE #233: [FE] Fix for solution submission and retrieval
# 📍Context 

During a migration in task #108  the route needed to be updated to appoint at new challenge microservice endpoint .
Accidentally creating the url, concatenating strings a `/` was missing causing an error when trying to submit or retrieve solutions. 

## 🔍Scope

- Avoid further delays in other tasks since the problem affects the usability of the product and the development process
- Prevent other errors in future calls to same endpont changing the string in environment

## ✅DoD
The get and post functions related to this service work and appoint to the correct endpoint

## ❌Out of Scope

- Changing the functions and how they manipulate response
- Create more environment variables for this endpoint

JungleGiu:
# 📍Context
In a previous PR  #741 
the migration of challenge submission and retrieval was executed but due to a typo in the route the calls returned a `404 -route not found` error .

## ✅What's done
The bug was fixed by adding a `/` in the same environment variable appointing to challenge user submission to avoid further problems in all requests involved

- `environment.ts`   `BACKEND_ITA_CHALLENGE_USER_SUBMISSIONS` variable
- `environment.prod.ts`  `BACKEND_ITA_CHALLENGE_USER_SUBMISSIONS` variable

## 🔍Proof

- Is it now possible to submit solutions to challenges
Screenshot 2026-02-22 alle 11 04 08

- The requests both return 200 status and appoint to the correct endpoint

Screenshot 2026-02-22 alle 11 03 44
Screenshot 2026-02-22 alle 11 04 02

- All tests pass

Screenshot 2026-02-22 alle 11 06 28


Arnau-66 (approved):
Changes look minimal and correct: updated the user submissions endpoint in both environments and added the corresponding changelog entry.
Everything looks good to me

Carlos-Martorell (approved):
Nice catch on the trailing slash! 

The fix looks good — applied correctly to both `environment.ts` and `environment.prod.ts` to keep them in sync.

Approved!

ivilarop (approved):

PR #775, opened 2026-02-23
PR 3/4 Update the frontend api client path to use the new get favorites endpoint and ensure no connections are lost (#202)

polserrano8:
📝 Description
Implementation of the frontend consumer migration for the Favorites GET endpoint path. This PR updates the frontend getUserFavorites() client call to use the new User-subresource contract (/user/users/{userId}/favorites) instead of the legacy userinteraction/favorites GET path.

This PR only changes the frontend GET favorites URL construction and the related unit test expectation. It does not change UI behavior, auth handling, Favorites business logic, or POST/DELETE favorites flows.
🔧 Technical Details
Path Change (implemented):

Legacy GET path:
GET /itachallenge/api/v1/userinteraction/favorites/{userId}

New GET path:
GET /itachallenge/api/v1/user/users/{userId}/favorites

Frontend Changes:
[challenge.service.ts](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/polse/.vscode/extensions/openai.chatgpt-0.4.76-win32-x64/webview/#):
Updated getUserFavorites() URL composition to the new endpoint path.

[challenge.service.spec.ts](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/polse/.vscode/extensions/openai.chatgpt-0.4.76-win32-x64/webview/#):
Updated the expected URL in the getUserFavorites() unit test to match the new route.

[environment.ts](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/polse/.vscode/extensions/openai.chatgpt-0.4.76-win32-x64/webview/#):
Updated the favorites route base constant from the legacy userinteraction path to the new /user/users base.

Config Alignment (required for production builds):
[environment.prod.ts](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/polse/.vscode/extensions/openai.chatgpt-0.4.76-win32-x64/webview/#) should expose the same constant name used by the service (BACKEND_USER_USERS_FAVORITES) with value /user/users.


src/environments/environment.prod.ts
```diff
@@ -5,7 +5,8 @@ export const environment = {
   appVersion: version,
   BACKEND_ITA_CHALLENGE_BASE_URL: '/itachallenge/api/v1',
   BACKEND_ITA_CHALLENGE_SOLUTION: '/challenge/solution',
-  BACKEND_ITA_CHALLENGE_USER_SUBMISSIONS: '/users/',
+  BACKEND_ITA_CHALLENGE_USER_SUBMISSIONS: '/users',
```
JungleGiu: This route it's out of your PR scope and was a fix to a solved bug I implemented this week successfully 
https://github.com/IT-Academy-BCN/ita-challenges-frontend/pull/771
Please change it back

src/environments/environment.ts
```diff
@@ -5,7 +5,8 @@ export const environment = {
   appVersion: version,
   BACKEND_ITA_CHALLENGE_BASE_URL: '/itachallenge/api/v1',
   BACKEND_ITA_CHALLENGE_SOLUTION: '/challenge/solution',
-  BACKEND_ITA_CHALLENGE_USER_SUBMISSIONS: '/users/',
```
JungleGiu, line 8: same here

JungleGiu (changes requested):

src/environments/environment.prod.ts
polserrano8: Sorry you're right my mistake!

src/environments/environment.ts
polserrano8: Again my mistake sorry!

polserrano8 (commented):
I've cleared the inconsistencies I had. Please review that the submission is correct now. Sorry for the rework!

JungleGiu (approved):
If the return type of the call is correct then everything is ok

zanlamar (changes requested):
Hello @polserrano8 ! 

✔️ All tests passing
✔️ All seem to be working locally after Giu's changes

🛑 Please check with backend collegues if versioning update of package.json, env. and changelog is needed. I think this change is relevant enough to add it. Even if afterwards the versioning will be changed, it is a good practice to always doing it and keeping it in mind.

🛑 Not a requested change, but I'd name shorter the branch and will not include the character # at the beggining. Check the project guidelines or other existing branches from previous iterations to ensure no problems with terminal checkout ignoring the branch due to # : https://tree.taiga.io/project/luisvi-ita-challenges/wiki/pr-taiga-worflow

Thank you!

JungleGiu (changes requested):
You should also update the versioning files according to documentation

PR #781, opened 2026-03-11
feat (challenge) deleteChallenge function added to challenge service #249

### LINKED ISSUE #249: [FE] Create deleteChallenge function in challenge Service
# 📍Context
The challenge service is the existing layer responsible for all API communication related to challenges. Currently there is no method to handle challenge deletion, so any component that needs to trigger this action has no service-level support to do so.

## 🏁 Objective
Create a deleteChallenge(id) method inside the challenge service that sends the appropriate HTTP DELETE request to the API and returns the response, making it reusable across any component that may need it.

## 🔧 Necessary Changes

- Add deleteChallenge(id: string) method to the challenge service
- Connect it to the correct API endpoint
- Handle and return both success and error responses properly
- Add any necessary types or interfaces if not already defined

## ❌ Out of Scope

- UI changes of any kind
- Calling or injecting the service into any component (covered in #252)
- Error handling at the component level
- Any changes to the API itself

JungleGiu:
# 📍Context
## [Related subissue](https://github.com/IT-Academy-BCN/ita-challenges/issues/249)
In this task [As a mentor, I can to delete a challenge]( https://github.com/IT-Academy-BCN/ita-challenges/issues/12)
we are aiming to create a new functionality for the authenticated mentor user, who could be capable of delete a challenge from the DB.
After considering the possible consequences in the way data is related the team decided to proceed and the function appointing to the existing endpoint `DELETE /itachallenge/api/v1/challenge/challenges/{challengeId}` was created in the challenge service

## ✅ What's done
- `deleteChallenge()` function was created in `challenge.service.ts` with appropriate error handling
- test coverage is been added for he new function

## 🔍 Proof
Screenshot 2026-03-11 alle 09 38 42

### Notes
The function is still not being called since this would be out of scope for this subissue, but the integration in the challengeForm component is next. to come in https://github.com/IT-Academy-BCN/ita-challenges/issues/252

> [!warning]
> During the solution of this issue a larger scope was identified to be necessary as the limitation of role must be managed by backend to guarantee edit and delete action can be executed only by users with an admin role. [Tech debt issue](https://github.com/IT-Academy-BCN/ita-challenges/issues/244) was produced and the role limitation will be handled securely frontend and backend side


otrocadev (approved):

polserrano8 (approved):
No he visto nada que me rechine en el código y la IA tampoco ha dado cosas a corregir que sean en este scope. Buen trabajo :)

src/app/services/challenge.service.ts
```diff
@@ -280,4 +280,21 @@ export class ChallengeService {
     )
   }
 
+  deleteChallenge(challengeId: string): Observable<any> {
+    const url = `${environment.BACKEND_ITA_CHALLENGE_BASE_URL}${environment.BACKEND_ALL_CHALLENGES_URL}/${challengeId}`
+    const headers = {
+      'Content-Type': 'application/json',
+      ...this.authService.getAuthHeaders()
```
ivilarop, line 287: The DELETE endpoint for challenges in back doesn't validate the user's role. An issue needs to be created to secure it. Have you discussed this with the back-end developers?

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [ita-challenges-frontend-3.35.0-RELEASE] - 2026-03-11
```
ivilarop: In the project, we're using UNRELEASED for functionality that isn't fully integrated yet. Can you update the CHANGELOG from RELEASE to UNRELEASED since the method isn't being called from any component yet?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ita-challenges-frontend",
-  "version": "3.34.0-RELEASE",
+  "version": "3.35.0-RELEASE",
```
ivilarop: leave the current version

conf/.env.CI.dev
```diff
@@ -1,2 +1,2 @@
 MICROSERVICE_DEPLOY=ita-challenges-frontend
-MICROSERVICE_VERSION=3.34.0-RELEASE
+MICROSERVICE_VERSION=3.35.0-RELEASE
```
ivilarop: leave the current version

ivilarop (changes requested):

CHANGELOG.md
JungleGiu: I'll change it right away

src/app/services/challenge.service.ts
otrocadev: Hello Ivan, we have talked about that. Our approach is that the user won't be able to use this endpoint if it is not a mentor. That's becaouse the form where the button for action will be is already in the same page as the form editing. So technically the approach is the same as now with thye editing permisions.

ivilarop: I understand the reasoning, but UI-based security isn't enough. Any user could directly call the DELETE endpoint without going through the form, for example, from Postman or curl. Role validation needs to be on the back end, regardless of what the front end does. Since the back end hasn't identified this need, it would be good to bring it to their attention so they can create a technical debt.

ivilarop (changes requested):

src/app/services/challenge.service.ts
JungleGiu: Yes we discussed the issue and since the ui of editing it's only accessible from a admin role, we talked about maintaining the same functionality. However I understand a more solid approach is recommended, I suggest to put a limitation in the next task , where we can verify the user role before triggering this function, that should be as abstract as possible in case of future backend role limitation implementations.

JungleGiu: Totally agree, i'll bring this issue to the attention of backend team

otrocadev: Should we also approach how this is handled with the edit process? I think the permissions and implications should be the same as the user role that shold be able to perform the action is the same.
 
We can see if this is already considered on the edit action of the same form so we can either repply the approach with the delete permission or to either make also a tech debt on this case scenario as well.

carlosPc1987:
I’ve added the tech debt in the link after reviewing the conversation. 



Technical debt created
Backend role validation for the DELETE challenge endpoint is now tracked as technical debt:

https://github.com/orgs/IT-Academy-BCN/projects/26/views/11?pane=issue&itemId=164773797&issue=IT-Academy-BCN%7Cita-challenges%7C263     

Once the version and CHANGELOG updates are done, this PR is good to merge from my side.

Great work, everyone

carlosPc1987 (approved):
good work ! approve

src/app/services/challenge.service.ts
ivilarop: In the pull request description, you should indicate the current problem: a user registered with the user role, if they switch to the user role, can edit/delete challenges and indicate the resulting technical issues. Regarding the technical issues, it should be specified that it is necessary to control, both front-end and back-end, that only users registered with the admin role in the database can edit/delete challenges.

ivilarop (changes requested):

src/app/services/challenge.service.ts
JungleGiu: Pull request updated to include tech debt and technical issues

ivilarop: Okay, thanks! Well done. When we deploy it, it would be good to explain to Luis during the demo that a student with user privileges can switch to admin and edit/delete challenges. You can close conversation

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [ita-challenges-frontend-3.35.0-UNRELEASED] - 2026-03-11
```
ivilarop: Shouldn't it be within a release section? In the last PR file we merge for this task, you should already include the release.

ivilarop (changes requested):

CHANGELOG.md
JungleGiu: Changes applied to changelog to unify to same release the last updates from the task

otrocadev: Till the last task https://github.com/IT-Academy-BCN/ita-challenges/issues/252, the full implementation of the feature will not be implemented working, should we update it to be a RELEASE before that?

ivilarop (approved):

PR #782, opened 2026-03-11
feature (shared) Created loading state enum #255

### LINKED ISSUE #255: [FE] Create a loading status enum with the options for loading, success and error scenarios
## Context
Currently, neither the challenge list (StarterComponent) nor the challenge detail view (ChallengeComponent) have a standardized way to represent the state of an async data fetch. Loading, success, and error states are handled implicitly (e.g., checking if an array is empty, or gating with *ngIf="challenge"), which makes it impossible to distinguish between "still loading" and "loaded but empty" or "failed".

> [!Important] 
> This is the first step of the parent task. It provides the shared foundation that all subsequent sub-issues (#256, #257, #258) depend on.

## Objective
Create a shared LoadingStatus enum that can be reused across any component that needs to track the lifecycle of an HTTP request.

## Scope
- Create a LoadingStatus enum in src/app/shared/enums/ with three values: LOADING, SUCCESS, ERROR.
- Export it so it can be imported by any component or service.

## Out of Scope
- No component logic changes — those are handled in #256 and #257.
- No UI/skeleton templates — those are handled in #258.

## Acceptance Criteria (DoD)
- [ ] A LoadingStatus enum exists in the shared enums folder with LOADING, SUCCESS, and ERROR values.
- [ ] The enum is properly exported and importable from other modules.
- [ ] Unit test covers that the enum has exactly the three expected values.

JungleGiu:
# 📍 Context
## [ Related subissue ](https://github.com/IT-Academy-BCN/ita-challenges/issues/255)
Since we are planning to implement a loading state  in the landing page [  (Parent Issue) ](https://github.com/IT-Academy-BCN/ita-challenges/issues/248) a unique standard for handling loading state is needed across the whole app. An enum seemed the way to avoid any inconsistencies in the future development of the app. 

## ✅ What's been done
- Created a `LoadingState` enum under `src/app/shared/enums/loading-state.enum.ts` with three values: `LOADING`, `SUCCESS`, and `ERROR`
- Exported the enum so it can be imported cleanly by any component or service
- Added a spec file (`loading-state.enum.spec.ts`) to verify the enum values and ensure no unintended changes are introduced in the future

## 🔍 Proof
Screenshot 2026-03-11 alle 17 22 02

###  🚩Notes
- The enum uses string values (e.g. `LOADING = 'LOADING'`) instead of numeric ones for better readability when debugging or logging
- This PR does not introduce any usage of the enum yet — adoption will happen incrementally in follow-up PRs as components and services are refactored 
- No breaking changes


otrocadev (approved):
Very simple and focused. Perfect. Let's keep going! 💪🏼

Lucy-SD (changes requested):
After reviewing the proposal, I'd only like to request a unified naming convention. We prefer to maintain `State` or align with `Status` ¿?

Regarding the test, a check could be added to prevent unexpected values, but this is not a blocking measure.

Another question I have is about the unreleased changes in the `Changelog`, which as I understand it, don't have a version number . . .

Almost perfect, nice and neat.

JungleGiu:
> After reviewing the proposal, I'd only like to request a unified naming convention. We prefer to maintain `State` or align with `Status` ¿?
> 
> Regarding the test, a check could be added to prevent unexpected values, but this is not a blocking measure.
> 
> Another question I have is about the unreleased changes in the `Changelog`, which as I understand it, don't have a version number . . .
> 
> Almost perfect, nice and neat

You're right a error was made redacting the documentation relative to this PR and the Status word was now changed to state to maintain a unified naming convention, thanks for noticing!

Lucy-SD (approved):

carlosPc1987 (approved):
Looks good to me!

The LoadingState enum is simple, well-structured, and provides a solid foundation for consistent loading state management across the app. Tests are appropriate and the CHANGELOG is properly updated. 

Approved.

carlosPc1987 (approved):
Well done! 

The LoadingState enum is clearly defined and properly tested. Changelog entry looks good too. Approved.

src/app/shared/enums/loading-state.enum.ts
```diff
@@ -0,0 +1,5 @@
+export enum LoadingState {
```
ivilarop, line 1: Reviewing the original code in `starter.component.ts`, we see that errors were handled solely with `console.error`, without any visual feedback to the user or a global mechanism (interceptor, toasting, etc.).

Given this context, the enum with three states (`LOADING`, `SUCCESS`, `ERROR`) adds an abstraction that goes beyond the scope of this task, whose objective was to implement a loading skeleton.

For this specific case, do you think a boolean would have been sufficient?

The enum would make sense if there were consistent visual error handling in the app, but introducing it here mixes two distinct responsibilities in the same task.

A good practice is to adjust the solution to the defined scope and open a separate issue if a major improvement is identified (such as global error handling). Did you have this in mind for a future task?

src/app/shared/enums/loading-state.enum.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  it('should have a SUCCESS value', () => {
+    expect(LoadingState.SUCCESS).toBe('SUCCESS');
+  });
+
+  it('should have a ERROR value', () => {
+    expect(LoadingState.ERROR).toBe('ERROR');
+  });
+
+  it('should have exactly 3 values', () => {
```
ivilarop: The `should have exactly 3 values` test is unnecessary—the three individual tests already guarantee that the values ​​exist and are correct. A test that validates the number of keys in an enum hinders extensibility without adding any real value.

ivilarop (changes requested):

src/app/shared/enums/loading-state.enum.ts
JungleGiu: I understand your concern about the scope of this task, we thought that as you say the no visual feedback in error handling could be addressed using this standard as good practice instead of a simple component level flag, so we decided to go in this direction.

I know it could seem like killing flies with a gun but I think it could give a path to follow for a next error and loading states implementation.
Tell me if you'd suggest to fall back on a boolean component level aprroach or if tech debt could be produced in this direction

src/app/shared/enums/loading-state.enum.spec.ts
JungleGiu: I'll delete this unnecessary test right away

src/app/shared/enums/loading-state.enum.ts
ivilarop: Great reasoning, and you're right that the enum provides a better foundation than a component-level flag if error handling is planned.

The key question is: **is there an open issue or plan for visual error handling in the app?** If yes, the enum is a solid forward-thinking decision. If not, it introduces an ERROR state in the template with a temporary message that no one has committed to designing properly — which is also a form of tech debt.

Both approaches carry trade-offs:
- A boolean is simple now but requires refactoring later if states grow
- An enum is extensible but adds complexity that only pays off if the ERROR state is actually used with proper UX

Given that, my suggestion would be: **keep the enum**, but remove the ERROR template from the `starter.component` for now and leave a TODO comment, until there's a proper issue for global error handling. That way you get the standard without shipping half-finished UI.

src/app/shared/enums/loading-state.enum.spec.ts
ivilarop: Okay, delete it and we'll resolve the conversation.

ivilarop (changes requested):

src/app/shared/enums/loading-state.enum.ts
JungleGiu: In the subissue https://github.com/IT-Academy-BCN/ita-challenges/issues/258 we are planning to address the UI of the loading skeleton and a different message to display to the user if an error in charging the data occurrs, since the actual message displayed is the 'No results' which is beeing kept to manage the absence of results after filtering. The messages displayed in https://github.com/IT-Academy-BCN/ita-challenges-frontend/pull/784 are just temporary as specified in the comments. 
However if you think the scope of the task needs to be reduced we are still on time (?) to do that

JungleGiu: We confronted to understand how to address this, but we thought if we don't implement an error message in case loading fails the user will never have clear feedback. Loading implies there are 2 different possible outputs from this state (Success or error). The error handling with console.error stays where it is, we are only adding a different visual feedback when the loading action fails to notify the user to try later or reload

ivilarop: ok thks! you can close

ivilarop (approved):

PR #780, opened 2026-03-10
Feat (challenge): Editor chellenge danger zone UI #250

otrocadev:
## 📍 Context
The editorChallenge component is where mentors manage challenge content. Currently there is no dedicated area for destructive actions. Mixing destructive actions with regular editing controls is a UX risk — a clearly separated Danger Zone section is a standard pattern to signal to the user that the actions within it are irreversible.

> [!Important] 
> This is one of the steps that are needed to accomplish the parent task https://github.com/IT-Academy-BCN/ita-challenges/issues/12

## 🏁 Objective
Add a visually distinct "Danger Zone" section at the bottom of the editorChallenge component that contains the delete trigger, making it clear to the mentor that this area holds destructive actions.

## 🔧 Changes Done
- Add a Danger Zone UI section to the editorChallenge component template
- Style it to be clearly distinguishable (e.g. red accent, warning label, visual separator)
- Place the delete action trigger (button) inside this section
- Ensure the section is only rendered for users with the appropriate mentor permissions

## ❌ Out of Scope
- The delete confirmation modal (covered in https://github.com/IT-Academy-BCN/ita-challenges/issues/251)
- Wiring up the actual delete function call (covered in https://github.com/IT-Academy-BCN/ita-challenges/issues/252)
- Any changes to the service layer
- Permission/role logic beyond consuming what already exists

## 🕹️ Demo
https://github.com/user-attachments/assets/cfbddce7-cfc3-4b7e-b1dd-8a41883744ac

PR #255, opened 2024-03-13
2.0.61-RELEASE feature#254

sacom14:
Delete itineraries.service.ts and move the method to challenge.service.ts

PR #783, opened 2026-03-12
Feature(challenge): Deleting challenge modal #251

### LINKED ISSUE #251: [FE] Create a modal for safe deletion, confirmation and error states
# PR https://github.com/IT-Academy-BCN/ita-challenges-frontend/pull/783

## 📍 Context
Deletion is an irreversible action. To prevent accidental data loss, a confirmation modal is required before the deletion is executed. The modal also needs to handle and communicate failure states clearly in case the API call does not succeed, so the mentor is never left without feedback.

## 🏁 Objective
Build a reusable confirmation modal component that handles three states: idle/confirmation, loading, and error — providing the mentor with a safe and informed deletion experience.

## 🔧 Necessary Changes

- Create a new modal component (or extend an existing one if available)
- Implement three UI states: confirmation prompt, loading/pending, and error message
- Confirmation action emits an event or calls a callback to proceed with deletion
- Cancel action closes the modal without triggering any changes
- Error state displays a clear message and allows the mentor to retry or dismiss

## ❌ Out of Scope

- The actual API call (covered in #249 and #252)
- Injecting or connecting the service inside this component
- Soft delete or undo functionality
- Any changes outside of the modal component itself

otrocadev:
## 📍 Context
Deletion is an irreversible action. To prevent accidental data loss, a confirmation modal is required before the deletion is executed. The modal also needs to handle and communicate failure states clearly in case the API call does not succeed, so the mentor is never left without feedback.

> [!important]
> This PR follows its previous one https://github.com/IT-Academy-BCN/ita-challenges-frontend/pull/780, that will be merged before.

## 🏁 Objective
Build a reusable confirmation modal component that handles three states: idle/confirmation, loading, and error — providing the mentor with a safe and informed deletion experience.

## 🔧 Changes done

- Create new modal components in the common modal service to manage all cases (Question, success and error)
- Implement three UI states: confirmation prompt, loading/pending, and error message
- Confirmation action emits an event or calls a callback to proceed with deletion
- Cancel action closes the modal without triggering any changes
- Error state displays a clear message and allows the mentor to retry or dismiss

## ❌ Out of Scope

- The actual API call (covered in https://github.com/IT-Academy-BCN/ita-challenges/issues/249 and https://github.com/IT-Academy-BCN/ita-challenges/issues/252)
- Injecting or connecting the service inside this component
- Soft delete or undo functionality
- Any changes outside of the modal component itself

## 🕹️ Demo

https://github.com/user-attachments/assets/6656d4e9-17c1-4c9f-8225-c7e0838ec58c


JungleGiu (approved):
Nice implementation using SweetAlert!  I think that's really clean and easy to adapt.
I only could review the code and not the functionality of it since server is down but everything looks good and ready to merge

Lucy-SD (changes requested):
After reviewing the PR and all related issues, I must say that the work is well implemented and correct.

Just a few questions . . .

* The changes to `challenge-form` seem out of scope. Is this so ¿? Should we move them to another PR, or what would be the best strategy ¿?

* Could the three modal methods be refactored to reduce duplication ¿?

* Would it make sense to use the `LoadingState` enum from #782 to maintain consistency ¿?

Overall, it's a great job, and I don't think any of this are blocking issues. I'm leaving it under "changes requested" just so we can analyze it and decide on the best strategy = ) 

Good work ¡!

src/app/services/challenge.service.ts
```diff
@@ -2,7 +2,7 @@
 /* eslint-disable @typescript-eslint/semi */
 import { Inject, Injectable, inject, signal } from '@angular/core'
 import { Observable, catchError, BehaviorSubject, of, throwError, forkJoin, switchMap } from 'rxjs'
-import { delay, map, tap } from 'rxjs/operators'
```
JungleGiu, line 5: was this an intended change?

otrocadev:
> After reviewing the PR and all related issues, I must say that the work is well implemented and correct.
> 
> Just a few questions . . .
> 
> * The changes to `challenge-form` seem out of scope. Is this so ¿? Should we move them to another PR, or what would be the best strategy ¿?
> * Could the three modal methods be refactored to reduce duplication ¿?
> * Would it make sense to use the `LoadingState` enum from #782 to maintain consistency ¿?
> 
> Overall, it's a great job, and I don't think any of this are blocking issues. I'm leaving it under "changes requested" just so we can analyze it and decide on the best strategy = )
> 
> Good work ¡!

Hello Lucy, thank you for asking and thank you for your review. About your questions and concerns:

- You are right about the fact that this PR should not change the ```challenge-form``` as it is something out of the scope. This is because this PR is heredating the changes from my previous PR. Maybe it is not well documented enaugh, so I am adding this in the description
- About the duplication, this is something that is demanded by the library SweetAlert2 that we are using, I have considered having the service of the deleting modals separated, but for now, I think it does necessarly make sense for the actual approach, as the project centralizes all alerts in the ```common-modal.service````
- About the LoadingStatus, I think it can make sense, depending on the feedback needed, and it is something we have thought about, to implement in the next user storie, where everything wraps together https://github.com/IT-Academy-BCN/ita-challenges/issues/252

src/app/services/challenge.service.ts
otrocadev: This is something I used for data mockup, and deleting the data I did delete it also, great job spotting it @JungleGiu ! Deleting it!

Lucy-SD:
> > After reviewing the PR and all related issues, I must say that the work is well implemented and correct.
> > Just a few questions . . .
> > 
> > * The changes to `challenge-form` seem out of scope. Is this so ¿? Should we move them to another PR, or what would be the best strategy ¿?
> > * Could the three modal methods be refactored to reduce duplication ¿?
> > * Would it make sense to use the `LoadingState` enum from [feature (shared) Created loading state enum #255 #782](https://github.com/IT-Academy-BCN/ita-challenges-frontend/pull/782) to maintain consistency ¿?
> > 
> > Overall, it's a great job, and I don't think any of this are blocking issues. I'm leaving it under "changes requested" just so we can analyze it and decide on the best strategy = )
> > Good work ¡!
> 
> Hello Lucy, thank you for asking and thank you for your review. About your questions and concerns:
> 
> * You are right about the fact that this PR should not change the `challenge-form` as it is something out of the scope. This is because this PR is heredating the changes from my previous PR. Maybe it is not well documented enaugh, so I am adding this in the description
> * About the duplication, this is something that is demanded by the library SweetAlert2 that we are using, I have considered having the service of the deleting modals separated, but for now, I think it does necessarly make sense for the actual approach, as the project centralizes all alerts in the ```common-modal.service````
> * About the LoadingStatus, I think it can make sense, depending on the feedback needed, and it is something we have thought about, to implement in the next user storie, where everything wraps together [[FE] Integrate service delete functionality with UI and modals ita-challenges#252](https://github.com/IT-Academy-BCN/ita-challenges/issues/252)

Thank you for all the clarifications and information regarding my doubts. As far as I'm concerned, they're resolved, and I believe your arguments are logical. Great job.

Lucy-SD (approved):

src/app/services/common-modal.service.ts
```diff
@@ -1,11 +1,12 @@
...
+import { inject, Injectable } from "@angular/core";
+import { TranslateService } from "@ngx-translate/core";
 import Swal, { SweetAlertOptions, SweetAlertResult } from "sweetalert2";
 
 @Injectable({
   providedIn: "root",
 })
 export class CommonModalService {
-  constructor() {}
+  private readonly translate = inject(TranslateService);
```
ivilarop, line 9: `constructor() {}` has been replaced with `private readonly translate = inject(TranslateService)`, which means that any spec that instantiates this service without providing `TranslateService` in the test module will fail.

I see that the service spec itself is already updated in this pull request. Have you verified that there aren't any other tests in the project that use `CommonModalService` directly and that might have broken with this change?

Are there other components in the app that use CommonModalService affected?

ivilarop (changes requested):

src/app/services/common-modal.service.ts
otrocadev: The empty constructor that was there before was just the default implementation for the service, but I removed it because it wasn't doing anything. I added the new injection to use the TranslateService that the new modals need to implement the translations correctly.
The translation though, is not being used in the rest of the modals, as they are hardcoded in Spanish for now, should I include this as a tech debt?

ivilarop: Thanks for the clarification! Yes, I'd suggest documenting it as tech debt — having some modals using `TranslateService` and others with hardcoded Spanish strings is inconsistent and should be addressed at some point.

That said, this actually makes the original question more pressing: if the existing modals never used `TranslateService`, their specs almost certainly don't provide it in the test module either. So the question remains — have you run `grep -r "CommonModalService" src --include="*.spec.ts"` to check if any other spec files instantiate the service directly? If they do, they'll now fail because `inject(TranslateService)` is required.

If the CI checks are all green, that's a good sign, but it would be great to confirm this explicitly before merging. 👍

ivilarop (changes requested):

src/app/services/common-modal.service.ts
otrocadev: I actually did verify this already. Here's what I found:

```grep -r "CommonModalService" src --include="*.spec.ts"``` returns 3 spec files, the testing checks are green on the three of them
image

The only spec that instantiates the real service is ```common-modal.service.spec.ts```, and it already provides TranslateService in the test module. CI is green and this is confirmed explicitly.

Agreed on documenting the hardcoded strings as tech debt, documenting it in the backlog.

To double check that this functionalities are working, I also reproduced them in my local front as a user. Here the proof:

https://github.com/user-attachments/assets/5926f3be-aeb4-45b9-bf2d-40806ee6d897


https://github.com/user-attachments/assets/213ef8e8-874a-42c6-a875-2591016c245a


I also saw that for now, there is no way for the mentor to create a new challenge via button or similar, the only way is to manually go to its specific rute, also adding this to tech debt.

ivilarop: Ok thks, you can resolve conversation

ivilarop (approved):

